PR #1, opened 2024-06-17
feat: Add a simple implementation of Provider Options.

douglasmakey:
First of all, thanks so much for this project; it's pretty amazing and also very useful. Currently, I am integrating it into a tool that solves some personal use cases, and I need to play with some fields for OpenAI like `temperature` and `max_tokens`. This is a tiny proposal to have the possibility to pass these values to specific providers. It's a pretty simple implementation.

Let me know what you think, and if the proposal is okay with you. If any extra work is needed, I will be happy to do it.

Thanks

### DIFF
--- a/examples/c04-provider-options.rs
+++ b/examples/c04-provider-options.rs
@@ -1,0 +1,45 @@
+use genai::chat::{ChatMessage, ChatRequest, ChatRequestOptions, OpenAIOptions, ProviderOptions};
+use genai::client::Client;
+use genai::utils::print_chat_stream;
+
+const MODEL: &str = "gpt-3.5-turbo";
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+	let questions = &[
+		// follow-up questions
+		"Why is the sky blue?",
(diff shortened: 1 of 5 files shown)


jeremychone:
@douglasmakey Thanks so much for the PR. I was planning to eventually add this, and I will bump it up in priority now.

I will implement this functionality, but with different implementations. The plan is to put those properties flat under ChatRequestOptions and have them generic for all adapters. Each adapter will be free to pick up what they can support.

In addition, I will probably have a cascading system where the ChatRequestProperties can be set at the client level (fallback for all adapter kinds and requests), at the adapter level for all requests of an adapter, and then at the execution level.


I think that future implementation should meet your requirement. Let me know if it does not. Later we can add more properties as it makes sense. 

We will maintain a table in the readme of that request option is supported for what adapter.

jeremychone:
@douglasmakey I implemented the feature and more in `v0.1.2` you can check the [examples/c04-chat-options.rs](https://github.com/jeremychone/rust-genai/blob/main/examples/c04-chat-options.rs) for more info.

PR #2, opened 2024-06-19
implement groq completions

stargazing-dino:
Implements groq api. See their [quickstart](https://console.groq.com/docs/quickstart)

Which for the most [part follows the openAI schema](https://console.groq.com/docs/openai).

One relatively larger change this PR made was to the `AdapterKind::from_model`.

The models from groq do not have any common prefix or suffix to do partial matching on so I had to change it to just match the exact string. For that I made the models pub. I thought to use `Adapter::list_models` but it required `kind` to be passed through which was unfeasible since we don't have one at that point and it's also async so it'd require some more changes than just making the models pub to the crate.

I tested this locally and it seemed to work fine.

### DIFF
--- a/examples/c00-readme.rs
+++ b/examples/c00-readme.rs
@@ -6,6 +6,7 @@ const MODEL_OPENAI: &str = "gpt-3.5-turbo";
 const MODEL_ANTHROPIC: &str = "claude-3-haiku-20240307";
 const MODEL_COHERE: &str = "command-light";
 const MODEL_GEMINI: &str = "gemini-1.5-flash-latest";
+const MODEL_GROQ: &str = "llama3-8b-8192";
 const MODEL_OLLAMA: &str = "mixtral";
 
 // NOTE: Those are the default env keys for each AI Provider type.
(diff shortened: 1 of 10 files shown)


jeremychone:
First, thanks so much for this PR.

The only thing I'm not a fan of right now is losing the "open" matching for the models that have clear prefixes.

For the from_mod, could we make it so that only Grok has the fixed name match, and the others remain as before?

If you do not have time to make the change, I will do it, no problem.

Having the fixed names for the list_models is okay for now. That will change later as I would like to have them query when the API is available.

Note: by design, the list_models is decoupled from the from_models, as the first one is more for user info and tries to be as live as possible, and the second is more for a wider match focused on in-memory/static processing.

stargazing-dino:
I moved back to the prefix solution :)

> Note: by design, the list_models is decoupled from the from_models, as the first one is more for user info and tries to be as live as possible, and the second is more for a wider match focused on in-memory/static processing.

ah makes sense

jeremychone:
Perfect, I will merge this. 

So Grok supports the OpenAI API to the letter? 
I found one bug in the Ollama OpenAI compatibility layer (it does not support multiple system messages), but I worked around it, so it all works. Hopefully, Grok does not have those types of issues.

stargazing-dino:
> So Grok supports the OpenAI API to the letter?

I think so, this is even their endpoint `https://api.groq.com/openai/v1/chat/completions`

> it does not support multiple system messages

I've tested groq and from what I remember it supports multiple system messages (for the llama models I tested at least) fine so no issue there hopefully

jeremychone:
Cool. I found this page:  https://console.groq.com/docs/openai

So it seems OpenAI API is their API strategy, so they should support it pretty well. 

I am going to add the temperature, max_token, … all in the ChatRequestOptions, so we will get that soon, and that will include Grok.

jeremychone:
@stargazing-dino Thanks, I just merged it.

jeremychone:
I would also concur with @Autumnlight02 that the experimentalDecorators need to be preserved. It's quite involved to move from one to another, so preserving both is very useful.

jeremychone:
Btw, very clean code and PR. Thanks!

PR #6, opened 2024-07-26
Add json mode for openai type models

stargazing-dino:
Support json_mode which forces model to respond with valid json

https://platform.openai.com/docs/guides/json-mode?lang=curl

https://console.groq.com/docs/api-reference#chat-create

### DIFF
--- a/src/adapter/adapters/openai/adapter_impl.rs
+++ b/src/adapter/adapters/openai/adapter_impl.rs
@@ -116,9 +116,11 @@ impl OpenAIAdapter {
 			payload["response_format"] = json!({"type": "json_object"});
 
 			// Check if "JSON" appears in the context
-			let json_in_context = messages
-				.iter()
-				.any(|msg| msg["content"].as_str().map_or(false, |content| content.contains("JSON")));
+			let json_in_context = messages.iter().any(|msg| {
+				msg["content"]
+					.as_str()
+					.map_or(false, |content| content.to_lowercase().contains("json"))
+			});
 
 			if !json_in_context {
 				return Err(Error::JsonModeWithoutInstruction);
--- a/src/adapter/adapters/openai/adapter_impl.rs
+++ b/src/adapter/adapters/openai/adapter_impl.rs
@@ -111,20 +111,9 @@ impl OpenAIAdapter {
 			"stream": stream
 		});
 
-		// -- Add options
+		// Add JSON mode if enabled
 		if let Some(true) = options_set.json_mode() {
-			payload["response_format"] = json!({"type": "json_object"});
-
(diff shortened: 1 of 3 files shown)


jeremychone:
Did a quick 15-second look at the commit, and it looks great. I will do a more thorough check, but this seems good to go. Thanks a lot.

jeremychone:
actually, why are we checking the user message content? It should just be a `json_output` kind of property in the `ChatOptions`

```rust
			let json_in_context = messages.iter().any(|msg| {
				msg["content"]
					.as_str()
					.map_or(false, |content| content.to_lowercase().contains("json"))
			});
```

jeremychone:
For now, I'm just trying to make sure I fully understand the behavior and purpose.

This check is for the list of messages before they are sent, right?

So, the goal of this behavior is to enforce that at least one of the messages has the word "json" in it to allow for a JSON call?

stargazing-dino:
Yeah, I don't think there's anything magic going on on their side either. As far as I understand `json_mode` just adds a validation check at the end of their ai response that verifies the response is valid json.

It will often fail if without proper prompting as well.

checking for the word "json" is just their way of ensuring the user is doing the bare minimum of prompting to ensure the model is guided to return json

jeremychone:
Ok, got it. However, I think this should be the responsibility of the caller to do this check, and it should be optional and independent of the `json_mode` (if/when we decide to implement this check).

So, I will merge your PR and then do another commit to remove that check.

But thanks, this is something I wanted to add sooner rather than later because I have the checkbox "JSON" in my AwesomeApp, but it was not wired to anything. So, good timing!

jeremychone:
Btw, I see why you did the check `"json"` in messages: 

(from the openai doc)

> Important: when using JSON mode, you must also instruct the model to produce JSON yourself via a system or user message. Without this, the model may generate an unending stream of whitespace until the generation reaches the token limit, resulting in a long-running and seemingly "stuck" request. Also note that the message content may be partially cut off if finish_reason="length", which indicates the generation exceeded max_tokens or the conversation exceeded the max context length.

So, as I said before, we might have a `json_mode_check` or something similar at some point. It's just tricky to do it correctly. It can only be a best effort (unless we assume we have a specialized agent to check it, which is out of scope).

That being said, for now, let's have this responsibility lie with the library user. I will put a doc note on the `json_mode` property about the need to do this check.

stargazing-dino:
Yes, that sounds reasonable

stargazing-dino:
I removed the extra logic and added a doc warning. Feel free to directly modify this PR further

jeremychone:
@stargazing-dino Ho, I already merged your commits, remove the json check. So, call good. This PR can be closed.

PR #9, opened 2024-08-02
fix: correct typos

upupnoah:


jeremychone:
@upupnoah Wow, that is a lot of typos. Thanks!

Will develop something to use LLMs for those in the future (Copilot is not that efficient at it).

Will merge now.

jeremychone:
@upupnoah big thanks for the clean PR, all merged.

PR #10, opened 2024-08-13
chore: make stream is send

giangndm:
This PR makes the stream with Send.

Why we need this: It allows us to use the stream inside a Tokio-spawned task without local_set.

### DIFF
--- a/src/chat/chat_stream.rs
+++ b/src/chat/chat_stream.rs
@@ -5,7 +5,7 @@ use futures::Stream;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
-type InterStreamType = Pin<Box<dyn Stream<Item = crate::Result<InterStreamEvent>>>>;
+type InterStreamType = Pin<Box<dyn Stream<Item = crate::Result<InterStreamEvent>> + Send>>;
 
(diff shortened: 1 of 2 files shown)


jeremychone:
@giangndm Thank you. Just merged it, will be in the next release. 

Btw, next release will change the way AuthResolver work. I will update the example and changelog. Ping me if any issue.

jeremychone:
@giangndm By the way, if you don't mind, could you update the README in your `genai-custom` to let developers know the official one is `genai`?

It's unfortunate that we can't delete crates, but perhaps you could yank your version as well to make it clearer.

giangndm:
@jeremychone Sure, I will update README for custom version and make it yanked after update my dependency

jeremychone:
@giangndm awesome, thanks!

jeremychone:
@giangndm I’m sure you forgot, but could you yank your genai-custom? It’s strange that somehow Google put the link above the official genai library. If you could yank all versions and maybe push another one with a new README that points to the crates.io genai, that would be great.

giangndm:
@jeremychone Sorry I have a very busy time. I just yanked the package and will update README and crate.io soon.

PR #12, opened 2024-08-17
make webc public

tusharmath:
The `error::Error` contains references to the `webc::Error` type


jeremychone:
@tusharmath Cool, good catch. 

FYI: I will merge this one, but add another commit to make sure only `webc::Error` get expose outside of the lib. 

Thanks

PR #14, opened 2024-08-17
add ci integration

tusharmath:


jeremychone:
At this point, the test requires API keys, and I am not comfortable putting my keys in an external system and having them used for PRs and such.

One could argue for doing a mock, but this would be a perfect example where mocks would not add much value. Things break in the details, and mocks will never be able to have those details (except with a tremendous amount of work, and even then, they will always play catch-up and require double the work).

So, for these reasons, I'll close this PR.

But @tusharmath, thank you very much for your contribution; it's really appreciated.

PR #15, opened 2024-08-17
chore: added feature flags for reqwest dep.

laststylebender14:


jeremychone:
@laststylebender14 

First, thank you for the PR.

I think it's a great idea to have a `with-rustls` flag.

The challenge is that this PR has too many commits to follow.

I am trying to follow the 1 commit per "intent" principle. I think that for this feature, everything should fit in one.

jeremychone:
I will close this PR so that it does not get confused with the eventual next one.

Issue #11, opened 2024-08-14
WASM support?


Boscop:
It would be nice to make the tokio dep optional so that it also works in wasm frontends :)

jeremychone:
I agree that this would be cool, but it could be too much to take on for now. 

I will close this one, but mark at as "might-be-for-later"

bachp:
I would be interested in WASM support, specifically wasip2 to be able to run genai on top of https://github.com/golemcloud/golem to build an Agent puerly in Rust.

As this was marked for mabye later almost 2 years ago, I wanted to check if this would be of interest now, before digging into what changes would be required.

jeremychone:
@bachp This would be a good capability to add, but I worry about the impact it would have on the code. 

For example, what library it might have to pull in, and the impact it would have on the main code by default.

Do you have some insight on those?

jeremychone:
Also, to be fully transparent, I am somewhat skeptical of WASM on the server. 

1) Native performance and capabilities for servers: After all the work on the Rust core, runtime, and ecosystem regarding performance and concurrency, introducing a middle-runtime for cloud processing has always looked counterproductive from my perspective, from a net value point of view. 

2) Running GenAI from a browser environment could be interesting, but having the keys on the client kind of goes against the best security practice of not trusting the (web browser) client with keys. So, typically, a web API to a server, and then the server handling this, would generally be the right approach for many production applications.

So, this is why regarding this capability, while it could be a popular feature, the fact that it could have a big impact on the codebase (design and implementation) and my perspective (which may be partial) on the subject, I did not spend much time thinking about this one.


Now, if this could just be an opt-in feature, with no change to the rest of the codebase, then I would be interested in considering it. 

But if it introduces changes in many core files with some sort of if/else WASM, then I will have to think hard about it.

bachp:
@jeremychone Thanks for the feedback. I just had a quick look over to code and I think the two main things that are not compatible with WASM are reqwest and tokio.

So what I think would be needed is the following refactorings:

1. Add a trait e.g. `HttpBackend` that abstracts over reqwest and use this in the library
2. Once reqwest is optional the only remaining usage of tokio seems to be in `printer` and `bedrock` so I would disable them for wasm

These refactorings should not change the behaviour so far. The next step would than be to add wasip2 support by:

3. Add new `HttpBackend` implementation based on some wasm compatible library e.g. [golem-wasi-http](https://github.com/golemcloud/golem-wasi-http)

[...]

PR #33, opened 2024-11-01
add: wasm target

tetuaoro:
Hi Monsieur Chone,

I don't know if these changes work because I'm poor xD. I'm starting a new project and I want to use this crate on the client side *(hydrate and csr)*.

### Changes

I used `target` instead of `feature` config to target *wasm* or *non-wasm* because some crates already implement *wasm* in their crate.

Thanks for this great library and happy to contribute to your project. Let me know what you think about my impl.

EDIT: related to #11

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,17 +16,27 @@ unused = { level = "allow", priority = -1 } # For exploratory dev.
 
 [dependencies]
 # -- Async
-tokio = { version = "1", features = ["full"] }
 futures = "0.3"
 tokio-stream = "0.1"
 # -- Json
 serde = { version = "1", features = ["derive", "rc"] } # Opted to rc for Arc<T> serialization
 serde_json = "1"
 # -- Web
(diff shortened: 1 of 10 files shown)


jeremychone:
@tetuaoro 

First, thanks for the contribution.

To be honest, I am a little worried about how intrusive the WASM support will be to the code. I understand the appeal, but not being an expert in Rust/WASM, I am going to be cautious on this one.

Do these changes work for you in WASM?

jeremychone:
First, thank you for your contribution.

I am going to close this PR, as I am not ready to add WASM yet. I need to better understand how intrusive it will be to the code.

Async/concurrency is critical to genai, so I am a little worried about how much change would be needed for WASM.

Also, I would probably put it behind a feature flag, and in most use cases, I do not see genai being used in a WASM context, except for some specific use cases.

So, closing this one for now, but will look at the code when/if I look into WASM. So thanks a lot.

PR #34, opened 2024-12-04
Add support for stop sequences to anthropic API

semtexzv:
Pretty self explanatory. See [Docs](https://docs.anthropic.com/en/api/complete)

### DIFF
--- a/src/adapter/adapters/anthropic/adapter_impl.rs
+++ b/src/adapter/adapters/anthropic/adapter_impl.rs
@@ -94,6 +94,10 @@ impl Adapter for AnthropicAdapter {
 			payload.x_insert("temperature", temperature)?;
 		}
 
+		if options_set.stop_sequences().len() > 0 {
+			payload.x_insert("stop_sequences", options_set.stop_sequences())?;
+		}
+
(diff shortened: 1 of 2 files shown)


jeremychone:
@semtexzv This is very cool. OpenAI supports that as well, so I will add it.

I will merge this commit, but afterward, I will make a change to use `Option<Vec..>`, and the getter will return an `Option<&[str]>`. Since the Stop Sequence will be rarely used, there's no need to initialize a vector for everyone.

Anyway, these are very minor details. Your PR is great (and just one commit, the way I like it). Thanks!

semtexzv:
> I will merge this commit, but afterward, I will make a change to use Option<Vec..>, and the getter will return an Option<&[str]>. Since the Stop Sequence will be rarely used, there's no need to initialize a vector for everyone.

Just FYI, empty vectors use dangling pointers. No cost to initialize them.

jeremychone:
@semtexzv Oh, you are completely right. I stand corrected.

A while back, I thought I read in some codebase that Option<Vec..> was a single word rather than three, and mistakenly extended from that. But even that was incorrect; they are both three words. I should have verified it. Thanks for the correction.

Since there's no need to differentiate between absent and empty, I think your solution is perfect.

PR #37, opened 2025-01-14
. fix links to c03 examples

bachp:
The example was renamed from c03-kind.rs to c03-mapper.rs

My editor also removed some trailing spaces. I hope that is OK.


jeremychone:
Thanks for the PR. I merged it. 

> It's a little unfortunate that the trailing spaces add noise, but perhaps I should have them cleaned as well. It would be nice if rustfmt could handle them. I will look into it.

PR #43, opened 2025-02-21
initial embedding implementation

GustavoWidman:
mostly boilerplate and basic embedding code for both openai and gemini providers. there are some things to fix, but this is currently working and the new tests (test_embed_single and test_embed_multiple) are passing for both gemini and openai providers, although the ModelIden function might have to be changed for embedding models, since it works based on prefix and embedding models aren't really standardized in terms of naming like completion models are.

### DIFF
--- a/src/adapter/adapter_kind.rs
+++ b/src/adapter/adapter_kind.rs
@@ -110,7 +110,10 @@ impl AdapterKind {
 			Ok(Self::OpenAI)
 		} else if anthropic::MODELS.contains(&model) || model.starts_with("claude") {
 			Ok(Self::Anthropic)
-		} else if cohere::MODELS.contains(&model) || model.starts_with("command") {
+		} else if cohere::MODELS.contains(&model)
+			|| cohere::EMBEDDING_MODELS.contains(&model)
+			|| model.starts_with("command")
+		{
 			Ok(Self::Cohere)
 		} else if gemini::MODELS.contains(&model)
 			|| gemini::EMBEDDING_MODELS.contains(&model)
--- a/src/adapter/adapter_kind.rs
+++ b/src/adapter/adapter_kind.rs
@@ -89,10 +89,10 @@ impl AdapterKind {
 	/// When more control is needed, the `ServiceTypeResolver` can be used
 	/// to map a model name to any adapter and endpoint.
 	///
-	///  - OpenAI     - starts_with "gpt", "o3", "o1", "chatgpt"
-	///  - Anthropic  - starts_with "claude"
-	///  - Cohere     - starts_with "command"
-	///  - Gemini     - starts_with "gemini"
+	///  - OpenAI     - model in OpenAI models or starts_with "gpt", "o3", "o1", "chatgpt"
+	///  - Anthropic  - model in Anthropic models or starts_with "claude"
+	///  - Cohere     - model in Cohere models or starts_with "command"
+	///  - Gemini     - model in Gemini models or starts_with "gemini"
 	///  - Groq       - model in Groq models
 	///  - DeepSeek   - model in DeepSeek models (deepseek.com)
 	///  - Ollama     - For anything else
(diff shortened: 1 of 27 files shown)


GustavoWidman:
let's keep this open for a while, this is nowhere near done

jeremychone:
This is a good way. Let's keep this PR open just to share code, but when/if we decide to implement it, let's open another clean PR with one commit and only the necessary changes.

jeremychone:
I just did a quick scan, and this is very clean code. Feel free to let me know when you think it is ready.

Also, it would be nice to have Nomic from Ollama; it's local, very fast, and a good one to have.

GustavoWidman:
> I just did a quick scan, and this is very clean code. Feel free to let me know when you think it is ready.
> 
> Also, it would be nice to have Nomic from Ollama; it's local, very fast, and a good one to have.

thanks. i might implement a few more providers this night, so i'll let you know. also, because ollama doesn't have openai compatible embedding api, i'll have to code the ollama provider separately, which is fine, but just means more work lol

jeremychone:
> thanks. i might implement a few more providers this night, so i'll let you know. also, because ollama doesn't have openai compatible embedding api (see https://ollama.com/blog/embedding-models), i'll have to code the ollama provider separately, which is fine, but just means more work lol

Yes, but same Ollama Adapter, right?

GustavoWidman:
> > thanks. i might implement a few more providers this night, so i'll let you know. also, because ollama doesn't have openai compatible embedding api (see https://ollama.com/blog/embedding-models), i'll have to code the ollama provider separately, which is fine, but just means more work lol
> 
> Yes, but same Ollama Adapter, right?

yes i can use the ollama adapter just fine, it's just a different endpoint and json format (which tbh can already be implemented easily given the current project structure)

GustavoWidman:
also, not sure what to do about the model_mapper mapping the embedding models incorrectly, do you have any ideas?

jeremychone:
@GustavoWidman why not do the same? If the user renames the wrong one, it will not work; otherwise, it will. No? Am I missing a use case?

GustavoWidman:
> @GustavoWidman why not do the same? If the user renames the wrong one, it will not work; otherwise, it will. No? Am I missing a use case?

im referring to the auto AdapterKind `from_model` function, which automatically extracts the AdapterKind from the given model

```
	pub fn from_model(model: &str) -> Result<Self> {
		if model.starts_with("gpt")
			|| model.starts_with("o3-")
			|| model.starts_with("o1-")
			|| model.starts_with("chatgpt")
		{
			Ok(Self::OpenAI)
		} else if model.starts_with("claude") {
			Ok(Self::Anthropic)
		} else if model.starts_with("command") {
			Ok(Self::Cohere)
		} else if model.starts_with("gemini") {
			Ok(Self::Gemini)
		} else if model.starts_with("grok") {
			Ok(Self::Xai)
		} else if deepseek::MODELS.contains(&model) {
			Ok(Self::DeepSeek)
		} else if groq::MODELS.contains(&model) {
			return Ok(Self::Groq);
		}
		// For now, fallback to Ollama
		else {
			Ok(Self::Ollama)
		}
	}
	
```

GustavoWidman:
there, nomic works! unfortunately, however, ollama does not support embedding dimension specification (like openai or gemini), so i guess we just have to ignore that client config :P

jeremychone:
@GustavoWidman Great work! I like your `from_model` approach; it will work for now.

I thought Nomic could support dimensions. I believe I used Ollama-rs a while back. I think I might have used it in my Ollama-rs video.

Obviously, the design of GenAI is not to use those per-provider crates, but we could look at how they did it.

GustavoWidman:
there, cohere works as well. the problem with cohere is that it introduces a totally new, required "input_type" field, which **has** to be passed to some models (v3 or above), see https://docs.cohere.com/v1/reference/embed

GustavoWidman:
> I thought Nomic could support dimensions. I believe I used Ollama-rs a while back. I think I might have used it in my Ollama-rs video.

will take a look, but im assuming they either dont use ollama directly or they redimension the vector themselves (spooky)

PR #44, opened 2025-03-02
Add tool use example

rgwood:
Hi, thanks for making `genai`; it's a really useful library!

This PR adds an example of tool use/function calling.

I was recently struggling to do tool use/function calling with `genai`. After I figured it out, I figured I should write up an example to save other people the time.


jeremychone:
@rgwood Merged. This is great, very clean PR.

I meant to do it but never took the time, so this is perfect!

Thanks a lot.

PR #48, opened 2025-04-07
Async auth

ehayes2000:
- Add support for async auth closures
- Add support for auth override url + headers
- Add examples

I'm using gemini through vertex and need to request a token before each request. These features are targeted at addressing this use-case (see `c09-google-gcp` for a concrete example of my use-case.

--- 
## Async auth
To avoid breaking changes to the API I've added async methods where appropriate. These methods could replace the existing methods as it's easy to wrap `FnOnce` in a future. 

I've marked several functions deprecated, as the config can be initialized with an async closure which cannot be used with the non-async functions. 

## AuthData::RequestOverride
This is pretty janky and was created to address more weirdness surrounding google auth. I think this functionality may better be split between the `service_target_resolver` and the `auth_resolver`, but I thought I'd ask before making larger changes.

--- 
These changes are based on the stable release.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,3 +36,4 @@ value-ext = "0.1.2"
 serial_test = "3.2.0"
 base64 = "0.22.0"  # Check for the latest version
 bitflags = "2.8.0"
+gcp_auth = "0.12.3"
--- a/examples/c08-contrived-async-auth.rs
+++ b/examples/c08-contrived-async-auth.rs
@@ -1,4 +1,4 @@
-//! This example demonstrates how to use a custom authentication function to override the default AuthData resolution
+//! This example demonstrates how to use a custom async authentication function to override the default AuthData resolution
 //! for any specific adapter (which is based on environment variables).
 
 use genai::chat::printer::print_chat_stream;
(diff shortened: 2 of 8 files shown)


jeremychone:
Ok, I am going to cose this one. Has confliec. I think I addrealy added the async auth.

ehayes2000:
Ahh I see what happened. My last PR was intended to be a PR for my own repo and should not have been merged it included some breaking changes which have been addressed in this PR. My mistake.

This PR also correctly uses `AsyncFnOnce` for async closures which makes the API much easier to consume as users can simple use a `async |model| Ok(Some(AuthData::...))` instead of the merged changes which requires consumers to use `|model| Box::pin(async move { Ok(Some(AuthData:...)) }}`

I think the breaking changes are less important as it is probably correct to use `async` functions for target and auth resolution, but It would be good to fix the closure issue.

I'll put up a Cherry-picked PR which addresses this soon. Sorry for the confusion.

jeremychone:
@ehayes2000 Ha, cool, thanks for the context. Yes, yesterday I had to do some work to make it work and not break the client sync/send behavior (which I used in my aipack cli).

I'm not super happy with what I came up with, but I needed to make it work. I changed it to `Fn`, and I didn't change the property `AsyncFnOnce` which I think would be better.

So if you can make another PR from main with the property `AsyncFnOnce` and update the `ServiceTargetResolver` as well, that would be great.

Then I'll test my AIPACK pass, and then I'll merge it.

I'm planning to do the official 2.0 release this week if all goes well.

Thanks again for the excellent work. Async closures are always a pain (which is getting better), and I'm very glad you helped tackle it.

Looking forward to your next PR. If it could be only ONE commit, that would be great.

PR #49, opened 2025-04-07
Async auth api (#2)

ehayes2000:
- Fix async closure api to allow `async || {...}` instead of `move | |  Box::pin(async {...})`

This should fix the api so any async closure can be moved. The changes here **are breaking** (they change some `Client` and `ClientConfig` to async methods), but from your merge of my previous PR it looks like this is ok.

### DIFF
--- a/examples/c08-contrived-async-auth.rs
+++ b/examples/c08-contrived-async-auth.rs
@@ -1,0 +1,53 @@
+//! This example demonstrates how to use a custom async authentication function to override the default AuthData resolution
+//! for any specific adapter (which is based on environment variables).
+
+use genai::chat::printer::print_chat_stream;
+use genai::chat::{ChatMessage, ChatRequest};
+use genai::resolver::{AuthData, AuthResolver};
+use genai::{Client, ModelIden};
+
+const MODEL: &str = "gpt-4o-mini";
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+	let questions = &[
+		// Follow-up questions
+		"Why is the sky blue?",
+		"Why is it red sometimes?",
+	];
+
+	// -- Build an auth_resolver and the AdapterConfig
+	// Auth function is captured with `from_resolver_async_fn` instead of the prior `from_resolver_fn`
+	let auth_resolver = AuthResolver::from_resolver_async_fn(
+		async |_model_iden: ModelIden| -> Result<Option<AuthData>, genai::resolver::Error> {
+			println!("Fetching auth!");
+			// this could be a network call
+			tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+			// This will cause it to fail if any model is not an OPEN_API_KEY
+			let key = std::env::var("OPENAI_API_KEY").map_err(|_| genai::resolver::Error::ApiKeyEnvNotFound {
+				env_name: "OPENAI_API_KEY".to_string(),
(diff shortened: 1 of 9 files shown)


jeremychone:
@ehayes2000 Quite a few changes, but great work. I love the simplicity of

```
	let auth_resolver = AuthResolver::from_resolver_async_fn(
		async |_model_iden: ModelIden| -> Result<Option<AuthData>, genai::resolver::Error> {
```

It's so simple now.

Thanks a lot. Let me test this on my side, and then I'll merge it if all is ok.

jeremychone:
@ehayes2000 It's breaking the client again, making it not `send`. Hopefully, it's only the `error::...` dyn that came back.

image

I'll fix it, or I won't change this PR; I'm working on it.

We have to have `client` be sendable. This is one of its main intended uses.

jeremychone:
I am almost there

```
impl<F: Send> AuthResolverAsyncFn for F
where
	F: AsyncFnOnce(ModelIden) -> Result<Option<AuthData>> + Send + Sync + Clone + 'static,
	F::CallOnceFuture: Send + 'static,
{
```

But weird, on the `F::CallOnceFuture` I get a "use of unstable library feature `async_fn_traits` " even it is supposed to be stable now. 

I set `rust-version = "1.86"`

Not sure what is happening.

jeremychone:
This is so unfortunate, `F::CallOnceFuture` is not stable yet :( 

So, if I accept this PR, then, the client, cannot be shared accross tasks/threads. 

And I just confirmed, the, with nightly and `#![feature(async_fn_traits)]` we are fine. 

This is so unfortunate.

ehayes2000:
I just ran into the same problem with the external error and fixed with 
```rust
External(Box<dyn std::error::Error + Send + Sync>),
```

The unstable `F::CallOnceFuture` is very sad. Maybe we could revert the traits to use `FnOnce`, this way async closures could be supported with the ugly api.

Thanks for reviewing. Let me know if there's anything else I can do to help this features move across the finish line.

jeremychone:
@ehayes2000 Unfortunately, I can't merge this one right now because `AsyncFnOnce::CallOnceFuture` isn't stable yet, even though `AsyncFn...` is.

I pushed a branch (xp-async-fn-once-resolvers) with this PR merged + some changes to fix what it broke, but it required `AsyncFnOnce::CallOnceFuture`, which is `#![feature(async_fn_traits)]` nightly (even with Rust 1.86).

https://github.com/jeremychone/rust-genai/tree/xp-async-fn-once-resolvers

jeremychone:
@ehayes2000 Yeah, right now, I think we're at a dead end. We have to use the ugly `Fn..` since `AsyncFn..` support isn't complete, even in 1.86.

As for the `External(Box<dyn std::error::Error + Send + Sync>)`, putting this as an error variant means the whole Error isn't sent, which then makes the client not send. That's why we have `Custom(String)`, which is good enough and async-friendly.

These things are hard to spot since we don't have any multi-task examples. I can easily see them on my AIPack since it fully uses it. I might make an example just to make sure we can easily spot when we break these things.

Anyway, great work; it won't be lost. We will use this work when full support is there. In a way, it was good to see it working on nightly. So, we have a working branch in a way.

Unfortunately, I'm not sure when Rust will add `AsyncFn...::CallOnceFuture` support. I guess it wasn't as simple as it looked, since they haven't done it yet.

PR #53, opened 2025-04-25
Feat: Add budget variant to ReasoningEffort

luna-spirito:
The `ReasoningEffort` enum now includes a `Budget(u32)` variant to support Gemini reasoning budget.

### DIFF
--- a/src/adapter/adapters/gemini/adapter_impl.rs
+++ b/src/adapter/adapters/gemini/adapter_impl.rs
@@ -3,7 +3,7 @@ use crate::adapter::gemini::GeminiStreamer;
 use crate::adapter::{Adapter, AdapterKind, ServiceType, WebRequestData};
 use crate::chat::{
 	ChatOptionsSet, ChatRequest, ChatResponse, ChatResponseFormat, ChatRole, ChatStream, ChatStreamResponse,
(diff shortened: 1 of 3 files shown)


jeremychone:
@luna-spirito Interesting what you did with `from_keyword`. So, this probably needs to be a `0.3.0` release.

I need to think about whether there are any other API changes I want to include, since we'll be doing a version bump.

jeremychone:
@luna-spirito FYI, I haven't forgotten about this PR.

I'll be adding it soon and will include more from/to string functions.

This will probably lead to a version bump to `0.3.0`, since the new variant might break some code. I'm trying to follow semver more strictly.

Note: For Gemini, I'm planning to add `-zero`, `-low`, `-medium`, and `-high` as alias suffixes, with `-zero` being a special one just for Gemini-only budget 0 (it won’t be a new variant).

jeremychone:
@luna-spirito Ok, I merged it and it’s now in master.

Sometimes GitHub doesn’t like the way I merge things.

But your commit is there, #f891e68fce2b7f018a4cec4c7b3b33328a0c6ec7, and I added you to the thank yous.

I added support for Gemini `-zero`, `-low`, `-medium`, and `-high` suffixes, and also mapped the other variants to the correct budget when present.

Big thanks, release 0.3.0 coming soon.

jeremychone:
v0.3.0 released!

PR #55, opened 2025-05-26
Update OpenAI adapter to check for tool calls if the LLM returns empty content

SilasMarvin:
I noticed that when using qwen3 with Ollama the LLM would often return empty content as well as tool calls.

I updated the OpenAI adapter to check if the content is empty before skipping the check for tool calls.

Also this is an awesome crate thank you!


jeremychone:
@SilasMarvin Great. Thanks for this tuneup. I am going to merge it as is. 

Thanks for the clean PR!

PR #56, opened 2025-06-06
Add Nebius adapter

AntonPates:


jeremychone:
@AntonPates First, thanks for this clean PR.

Now, I have mixed feelings about this. Should it be done with a custom endpoint using the `ServiceTargetResolver` or should it have its own adapter?

It seems like Nebius is not very well known.

Maybe having a convenient `ServiceTargetResolver` builder that takes a list of model names, an endpoint, and env for keys would be enough?

I'm just curious and debating with myself. For example, I think I will add an `OpenRouter` adapter at some point (it was requested), but sometimes some specific ones might be better done via Resolver. I am not sure.

I am definitely okay with adding this, just validating and maturing the "when to make it an adapter" strategy.

What do you think?

jeremychone:
@AntonPates Thanks for the context...

Your PR/Code is perfect, just that given genai's current design for Adapter mapping from model name only (by default), those "model names" cloud be used for other Adapters/Contexts, and then the user might get confused.

One thing I was thinking for model names is to introduce the concept of "namespacing," which will basically be the adapter name in lowercase.

So,

`nebius::Qwen/Qwen3-235B-A22B`

This way, we add the adapter, nebius. But it's only accessible when the model name is namespaced.

This will also be implemented for all other adapter kinds, so one could do `ollama::Qwen/Qwen3-235B-A22B`, and assuming Ollama has this, it will work (note: typically Ollama does not use `/` but this was just to illustrate).

Would that work for you?

(Do not make any changes for now. I might just merge your commit, add the "model namespace feature," and then remove the MODELS name matched for nebius, and it will just work with `nebius::`)

AntonPates:
@jeremychone I believe namespaces would be a convenient way to avoid confusion anticipating growing number of adapters

jeremychone:
@AntonPates I merged this PR, added the "model namespace" feature, and removed the model name mapping since now you can do:

`nebius::Qwen/Qwen3-32B` (just prefixed with the `nebius::` namespace)

This works for all adapters. For example, if we have an OpenAI `agi-5` model, the mapping isn’t updated, and users can use the OpenAI adapter with `openai::agi-5`

As usual, for more advanced use cases, we have the `ServiceTargetResolver`.

Let me know if this solution doesn’t work for you. Thanks a lot for the PR!

PR #58, opened 2025-06-12
Add Tool Use Streaming Support and c10-tooluse-streaming Example

ClanceyLu:
This PR introduces streaming support for tool use and provides a working example under examples/c10-tooluse-streaming.

### DIFF
--- a/src/adapter/adapters/gemini/adapter_impl.rs
+++ b/src/adapter/adapters/gemini/adapter_impl.rs
@@ -18,8 +18,8 @@ pub struct GeminiAdapter;
 const MODELS: &[&str] = &[
 	"gemini-2.0-flash",
 	"gemini-2.0-flash-lite",
-	"gemini-2.5-flash-preview-04-17",
-	"gemini-2.5-pro-preview-05-06",
+	"gemini-2.5-flash-preview-05-20",
+	"gemini-2.5-pro-preview-06-05",
 	"gemini-1.5-pro",
 ];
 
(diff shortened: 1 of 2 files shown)


jeremychone:
@ClanceyLu Thanks. This is a big one. There was one that did that a while back, but it broke other things. So, I will check this one.

Now, I don't understand the `.gemini model names ...` commit. This is a change I made a while back, so why is it in this PR?

jeremychone:
@ClanceyLu Ok, on a first scan, this looks very well made and follows everything.

Just curious, why is `capture_tool_calls` an option?

```
	/// (for streaming only) Capture all tool calls from tool call chunks
	/// `StreamEnd` from `StreamEvent::End(StreamEnd)` will contain `StreamEnd.captured_tool_calls`
	pub capture_tool_calls: Option<bool>,
```

I did that for a reason, since some or most users might not care about it.

But for tool_calls, what would be a use case where someone wouldn't care about it? If I do not capture it, then the response is incomplete and not coherent.

ClanceyLu:
You’re absolutely right — if tool_calls aren’t captured, the response would indeed be incomplete unless the user handles ChatStreamEvent::ToolCallChunk manually.

That’s actually why I made capture_tool_calls an Option<bool> — to keep things flexible for advanced users who may already be aggregating tool calls from the stream themselves. Setting this option to true just gives them a ready-to-use captured_tool_calls at the end of the stream, without taking away control from users who want to process chunks incrementally.

Happy to revisit this if a default true behavior makes more sense going forward.

jeremychone:
@ClanceyLu Oh, you are right about the `capture_tool_calls`. I was confusing myself.

Yes, when doing the stream call, users get the tool calls as part of the stream. If they want the convenience of having it at the end as well, they use `capture_tool_calls`.

This is perfect!!!

Great work!

PR #67, opened 2025-07-01
feat: support built-in provider tools such as Google Search for Gemini

4t145:
This pr would allow us to use "embeded" tools like google search, witch will be called on server side instead of client side.
# Example

```rust
        llm
            .exec_chat(
                &config.query_generator_model,
                ChatRequest::new(vec![ChatMessage::user(
                    WebSearch {
                        research_topic: &query.query,
                        current_date: &chrono::Utc::now().to_rfc3339(),
                    }
                    .format_prompt(),
                )])
                .with_tools(vec![
                    Tool::new("googleSearch").with_config(serde_json::json!({})),
                ]),
                Some(&ChatOptions::default().with_temperature(0.0)),
            )
            .await?;
```

### DIFF
--- a/src/adapter/adapters/gemini/adapter_impl.rs
+++ b/src/adapter/adapters/gemini/adapter_impl.rs
@@ -150,12 +150,7 @@ impl Adapter for GeminiAdapter {
 
 		// -- Tools
 		if let Some(tools) = tools {
-			payload.x_insert(
-				"tools",
-				json!({
-					"function_declarations": tools
-				}),
-			)?;
+			payload.x_insert("tools", tools)?;
 		}
 
 		// -- Response Format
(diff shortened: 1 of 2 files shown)


jeremychone:
@4t145 
This is an interesting one, but there are a few things that worry me.

1) The tool names are hardcoded in the adapter. There might not be another way, and this might be ok since it is only for that adapter.

2) I do not get the `Tool Config`, and then we have `tool.name: tool.config`. Does a Google tool name expect a struct? This behavior is not really intuitive, putting a config in a name.

3) Also, I noticed another change, going from

```
			payload.x_insert(
				"tools",
				json!({
					"function_declarations": tools
				}),
			)?;
```
to
```
payload.x_insert("tools", tools)?;
```
Did I have a bug before? (I might have, since it was snake case.)

---

Anyway, this is an interesting PR. I just wish it would have been possible to do this with just standard tools. I need to play with it a little.

PR #68, opened 2025-07-01
feat: remain original response body

4t145:
Remain original response body, so users can do provider specific operations.

### DIFF
--- a/src/adapter/adapter_types.rs
+++ b/src/adapter/adapter_types.rs
@@ -32,6 +32,7 @@ pub trait Adapter {
 
 	/// To be implemented by Adapters.
 	fn to_chat_response(
+		client: &crate::Client,
 		model_iden: ModelIden,
 		web_response: WebResponse,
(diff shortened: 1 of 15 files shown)


jeremychone:
@4t145 so, you want the option b) or a) ?

4t145:
@jeremychone Now you can review again.

jeremychone:
@4t145 Cool, will look at it tomorrow or Friday.

jeremychone:
@4t145 Ok, overall looks good. One thing, is that you added "client" to the `to_chat_response` to get th options, but we alreayd have the `option_sets`. 

I will merge this PR and fix it. 

Also, the property to get the content should be `captured_raw_body`. 

But otherwise, pretty good. I like the way you did `let capture_raw_body = client.config().capture_raw_body().then(|| body.clone());`

jeremychone:
I understand the confusion. You put it in ClientConfig, which is understandable. But we want this at the Chat Options level so hat we can override it per chat.

PR #69, opened 2025-07-03
fix: improve reasoning content extraction in OpenAIAdapter

XciD:
Updated the reasoning content extraction logic to handle cases where reasoning may be present in multiple locations within the response. Enhanced error handling for the extraction process to ensure robustness.


jeremychone:
@XciD Thanks for the very clean PR. This is a good fix. Thanks!

PR #71, opened 2025-07-03
Extend OpenAI adapter to preserve request params

hobofan:
This change adjusts the way the service_url for the OpenAI adapter is built, so that Azure OpenAI endpoints can be supported.

As Azure OpenAI requires an `api-version` parameter to be set, a simple way to provide them would usually be to set a base URL for the providers, that ends with something like `?api-version=2024-10-08`. However, if we then use the naive current way to build the full service URL, these params end up in the middle of the URL. (Sadly a requwest client doesn't have an alternative extension method like `default_parameters` like it does for headers with `default_headers`).

With this change, the base_url, is first parsed to extract the params, and after building the full url, they are added back in.


jeremychone:
Perfect, thanks!

PR #75, opened 2025-07-08
keep header in error

XciD:
usefull to display debug headers in case of error


jeremychone:
Ok, make sense. 

Eventually, those errors might become "Json Serializable," but we'll deal with that if/when it happens.

PR #76, opened 2025-07-09
Implement zhipu (ChatGLM) completions 

zilongpa:
This PR adds support for Zhipu (Z.ai) models, which closely follow the OpenAI API schema.
All models are prefixed with `glm` and have been verified locally to work correctly.

Notable Differences from OpenAI:
- **No implicit caching**: Each request is evaluated fully without automatic reuse of prior responses.
- **No native structured output**: The model does not support returning strict JSON.
- **Stop tokens are included**: The stop word is returned as part of the generated text (as the final token).

### DIFF
--- a/examples/c00-readme.rs
+++ b/examples/c00-readme.rs
@@ -12,6 +12,7 @@ const MODEL_GROQ: &str = "llama-3.1-8b-instant";
 const MODEL_OLLAMA: &str = "gemma:2b"; // sh: `ollama pull gemma:2b`
 const MODEL_XAI: &str = "grok-beta";
 const MODEL_DEEPSEEK: &str = "deepseek-chat";
+const MODEL_ZHIPU: &str = "glm-4-plus";
 
 // NOTE: These are the default environment keys for each AI Adapter Type.
 //       They can be customized; see `examples/c02-auth.rs`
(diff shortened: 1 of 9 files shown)


jeremychone:
Hi @zilongpa , I merged a PR refactoring Zhipu to `zai`, which makes sense since it points to the `z.ai` API site.

It’s a little confusing, though, since we currently use Zhipu under `bigmodel.cn`.

We can add another adapter to `bigmodel.cn` called `bigmodel`. This way, we can have the namespace `bigmodel::glm...`

Let me know.

jeremychone:
See more conversation there: https://github.com/jeremychone/rust-genai/pull/76#issuecomment-3594311524

tomsun28:
> @zilongpa @tomsun28 @shingonoide
> 
> FYI all, change in the zai adapter coming in `0.5.0`
> 
> I realized that the "zai" adapter has this dual endpoint, which is kind of ok, but I think the way it is implemented is wrong.
> 
> If there is a namespace, it goes to coding, and if there is not, it goes to regular.
> 
> But this is not the way namespaces in genai should work. Doing a `namespace::..` should be the same as if the model was selected by default.
> 
> But I get the feature. So, here is how it will work (I will make the change):
> 
> * if default `glm..` model name or `zai::...` will use the regular endpoint.
> * and I will introduce a `zai-coding::` namespace, and this will use the other one. This way, it is cleaner and more predictable.
> 
> Hope it makes sense. Feedback welcome.

@jeremychone hi thanks for informing us. I think the new namespace `zai-coding` is a better design. 👍

jeremychone:
@tomsun28 @zilongpa Btw, for the zai subscription, for `v0.6.0` it will be `zai_coding`, the previous one was `coding` which is why too generic.

PR #78, opened 2025-07-10
feat: add extra headers to requests

XciD:
Allow user to add some extra headers to inference requests

### DIFF
--- a/src/adapter/adapters/openai/adapter_impl.rs
+++ b/src/adapter/adapters/openai/adapter_impl.rs
@@ -194,10 +194,15 @@ impl OpenAIAdapter {
 		let url = AdapterDispatcher::get_service_url(&model, service_type, endpoint);
 
 		// -- headers
-		let headers = vec![
+		let mut headers = vec![
 			// headers
 			("Authorization".to_string(), format!("Bearer {api_key}")),
 		];
+
+		// -- extra headers
+		if let Some(extra_headers) = options_set.extra_headers() {
+			headers.extend(extra_headers.iter().cloned());
+		}
 
 		let stream = matches!(service_type, ServiceType::ChatStream);
 
--- a/src/chat/chat_options.rs
+++ b/src/chat/chat_options.rs
@@ -64,6 +64,9 @@ pub struct ChatOptions {
 	/// Set the seed
 	/// This is useful for reproducibility.
 	pub seed: Option<u64>,
+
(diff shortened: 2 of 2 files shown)


jeremychone:
perhaps two function
.headers_override(vec..
.headers_append(vec..

not sure, just brainstorming.

jeremychone:
Btw, I am going to release a 0.4.0-alpha.4 now. We have enough new things.

jeremychone:
I use the `.header...` method from reqwest, which probably, in turn, uses `req.headers_mut().append(key, value);`

```rust
	pub fn new_req_builder(&self, url: &str, headers: &[(String, String)], content: Value) -> Result<RequestBuilder> {
		let method = Method::POST;

		let mut reqwest_builder = self.reqwest_client.request(method, url);
		for (k, v) in headers.iter() {
			reqwest_builder = reqwest_builder.header(k, v);
		}
		reqwest_builder = reqwest_builder.json(&content);

		Ok(reqwest_builder)
	}
```

So maybe in the `new_req_builder` function, we should trim the headers so the last name/value pair takes precedence.

What do you think?

Then, later on, if we need multi-value headers, we can offer another API.

jeremychone:
@XciD Yes, I think it's the right time to switch to a Map. It better conveys the intent.

> Note: Now I remember, I had decided to use vec previously to allow for full header functionality.
> However, thinking more about it, I believe it creates more issues than it solves. We can offer lower-level APIs later, and staying with vec<..> would be misleading.

So, I'm going to make this change.

jeremychone:
@XciD Okay, it's been added: 1c69d18224742461a143b4a4b3f1b6ab28ef62b9

I created a new type, `Headers`, which, for now, only contains a `HashMap<String, String>`, along with all of the useful `From`, `merge`, and `merge_with` implementations.

So, it's an API change, but definitely for the better, and the timing is good with the upcoming 0.4.0 release.

I'm also going to add it to `ServiceTarget`, which will be another API change, but allow users to completely control the end headers.

jeremychone:
@XciD FYI, I just released 0.4.0 with all of the new features: PDF, Images, custom headers, namespace (fireworks::gpt-oss-20b), a more streamlined MessageContent, and many fixes.

Thanks a lot for your contribution to this release.

Feel free to ping me if you have any issues or suggestions for the future.

PR #79, opened 2025-07-10
Implement support for custom HTTP headers in chat completion calls

julien-c:
> [!CAUTION]  
> ⚠️ Disclaimer: this PR was loosely (given i'm a rust noob) humanly reviewed  but is otherwise mostly AI-generated
> ~ julien-c

This PR implements the ability to add custom HTTP headers to chat completion calls, addressing the need for users to include additional metadata, authentication tokens, or tracking information in their API requests.

### Core Implementation
- **Added `extra_headers` field** to `ChatOptions` struct as `Vec<(String, String)>`
- **Added chainable setter methods**:
  - `with_extra_header(key, value)` for adding individual headers
  - `with_extra_headers(headers)` for adding multiple headers at once
- **Updated `ChatOptionsSet`** to handle extra headers with proper cascading (chat-level overrides client-level)

### Adapter Updates
- **Updated all adapter implementations** to include extra headers in HTTP requests:
  - OpenAI adapter (via utility function - also applies to Groq, xAI, Ollama, DeepSeek, Zhipu, Nebius)
  - Anthropic adapter
  - Gemini adapter  
  - Cohere adapter

### Usage Examples

```rust
use genai::{Client, ChatOptions, ChatRequest, ChatMessage};

// Add individual headers
let options = ChatOptions::default()
    .with_extra_header("X-Custom-App", "my-rust-app")
    .with_extra_header("X-Request-ID", "req-12345");

// Add multiple headers at once
let headers = vec![
    ("X-Custom-App".to_string(), "my-rust-app".to_string()),
    ("X-Request-ID".to_string(), "req-67890".to_string()),
];
let options = ChatOptions::default().with_extra_headers(headers);

// Use in chat calls
let client = Client::default();
let chat_req = ChatRequest::from_user("Hello!");
let response = client.exec_chat("gpt-4o-mini", chat_req, Some(&options)).await?;
```

## Testing
- **Unit tests** for `ChatOptions` and `ChatOptionsSet` functionality
- **Integration tests** for OpenAI adapter to verify headers are properly included in `WebRequestData`
- **Example application** (`examples/c11-extra-headers.rs`) demonstrating usage

## Compatibility
- Fully backward compatible - existing code continues to work unchanged
- Headers follow the same cascading pattern as other chat options
- Empty headers list by default (no behavior change when not used)

### DIFF
--- a/examples/c11-extra-headers.rs
+++ b/examples/c11-extra-headers.rs
@@ -1,0 +1,44 @@
+//! Example demonstrating how to add extra headers to chat completion calls
+
+use genai::Client;
+use genai::chat::{ChatMessage, ChatRequest, ChatOptions};
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Create a client
+    let _client = Client::default();
+
+    // Create a simple chat request
+    let _chat_req = ChatRequest::new(vec![
+        ChatMessage::system("You are a helpful assistant."),
+        ChatMessage::user("Hello, how are you?"),
+    ]);
+
+    // Example 1: Add extra headers using individual with_extra_header calls
+    let options_individual = ChatOptions::default()
+        .with_extra_header("X-Custom-App", "my-rust-app")
+        .with_extra_header("X-Request-ID", "req-12345")
+        .with_temperature(0.7);
+
+    // Example 2: Add extra headers using with_extra_headers with a vec
+    let headers = vec![
+        ("X-Custom-App".to_string(), "my-rust-app".to_string()),
+        ("X-Request-ID".to_string(), "req-67890".to_string()),
+        ("X-User-Agent".to_string(), "GenAI-Rust-Client/1.0".to_string()),
+    ];
+    
+    let options_batch = ChatOptions::default()
+        .with_extra_headers(headers)
+        .with_temperature(0.7);
+
+    println!("Extra headers example - these would be included in HTTP requests:");
+    println!("Individual headers: {:?}", options_individual.extra_headers);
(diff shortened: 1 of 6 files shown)


julien-c:
@copilot please fix the merge conflicts on main

jeremychone:
@julien-c Yes, that was interesting, and thanks so much for the disclaimer, very much appreciated.

It was interesting to see the difference between Copilot's work versus a human coder's (#78).

If we look at #78, it was a very clean PR for a simple feature: no noise, a single commit, just straight to the point. This made it super efficient for me to review, and I even clicked "squash and merge" from the UI (which I typically don't like to do).

This Copilot PR was the opposite. It had a whole plan and three commits for such a simple feature. I understand why Copilot generally did it this way, but this is where human control adds a lot of value.

This aligns with my past experience that fully generated PRs tend to have noisier, lower quality output than human-controlled output (with or without AI assistance).

This is why I tend to prefer new Rust developers to log requirements tickets rather than asking Copilot for the feature. I can use AI more effectively, and it will save more time for all parties.

Now, don't get me wrong, I definitely think developers should and even must use AI for production coding as much as possible; however, it has to be micromanaged, and the end result needs to be human-owned.

This is why I created [AIPACK](https://aipack.ai) with [pro@coder pack](https://news.aipack.ai/p/production-coding-example-with-procoder), and I code AIPACK with AIPACK (about 40% code is generated by AI, but micro-managed by me). eventually, I can even have an agent help understand the library and patterns, to help developers making PRs using the lib best practices, but in the end, it has to be developer-managed end-to-end.

Anyway, this was a very interesting experiment, and we have a clear comparison.

julien-c:
awesome thanks, cc @XciD

PR #80, opened 2025-07-14
Add support for tool calls and thinking in the Anthropic streamer.

AndrewRademacher:
Add support in the Anthropic Streamer for:

* Tool Call Deltas
* Thinking Deltas

I was not able to find a generic streaming test with tool calls in the repo, though I may be blind. If there is one that I have missed I'll add it to the Anthropic tests. Otherwise, I can infer what one should be.


jeremychone:
@AndrewRademacher Thanks for the clean PR. I ran the stream test, and it worked.

> Note for later: Eventually, we'll need to add a stream test function.

PR #82, opened 2025-07-16
Add anthropic PDF support.

AndrewRademacher:
This is my best guess at PDF and other document support that is in line with how the rest of the library is structured. Other providers are currently no-oped. Does this fit with your expectation of how documents should work in this library?

There is still a rustfmt issue here, looks like it didn't see that you have a custom config. I'm looking into it.

### DIFF
--- a/src/adapter/adapters/anthropic/adapter_impl.rs
+++ b/src/adapter/adapters/anthropic/adapter_impl.rs
@@ -332,12 +332,12 @@ impl AnthropicAdapter {
 												"url": url,
 											}
 										})),
-										DocumentSource::Base64(bytes) => Some(json!({
+										DocumentSource::Base64 { file_name: _, content } => Some(json!({
 											"type": "document",
 											"source": {
 												"type": "base64",
 												"media_type": "application/pdf",
-												"data": bytes,
+												"data": content,
 											}
 										})),
 									},
--- a/src/adapter/adapters/anthropic/adapter_impl.rs
+++ b/src/adapter/adapters/anthropic/adapter_impl.rs
@@ -1,17 +1,17 @@
-use crate::ModelIden;
 use crate::adapter::adapters::support::get_api_key;
 use crate::adapter::anthropic::AnthropicStreamer;
 use crate::adapter::{Adapter, AdapterKind, ServiceType, WebRequestData};
 use crate::chat::{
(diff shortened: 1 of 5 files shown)


jeremychone:
@AndrewRademacher 
First, thank you for working on this.

A few things.

First, it's very important that the official pull request (PR) to be merged has only one commit (the rustfmt.toml is in the repo). Otherwise, it's very hard for me to mentally reconcile the various commits, and I'd need to pull it and squash them into a local branch, then check them there. Not impossible, but it's more work and might cause delays, leading to future conflicts.

Second, I might change a couple of things about how we handle multipart. We currently have `MessageContent`, which is a list, and then parts with their content. I might flatten the content part into `MessageContent` to have:

```rust
MessageContent::Text(String),
MessageContent::Image(content_type, source: base64String) // (might become Image or Binary{content_type, source: base64}),
MessageContent::ToolCalls //  (still `Vec<ToolCall>`)
MessageContent::ToolResponses //  (still `Vec<ToolResponse>`)
```

This is because `content` is already a `Vec<MessageContent>`.

So, I guess PDF support will be impacted by that. Perhaps it will require another type (`MessageContent::PDF`).

Anyway, please make sure the entire PR is one commit. Feel free to close this one and create another when you're ready. Then, I will decide whether to do this refactoring before or after the PR is merged.

Does that make sense? Am I missing anything?

jeremychone:
@AndrewRademacher Give me the weekend to make the change on the MessageContent side. Then, we can discuss what is missing for the Pdf support.

jeremychone:
@AndrewRademacher I implemented PDF support using ContentPart::Binary as we discussed, and I kept the ContentPart for now.

By the way, thank you—I reviewed your PR.

I released version 0.4.0-alpha.10.

I also added fireworks:: and together:: adapters or namespaces, and I’m particularly impressed with Fireworks and some of those open-source models.

Next, I plan to explore flattening ContentPart into a ContentMessageList or something similar.

I wanted to release alpha.10 with the initial version of PDF support.

jeremychone:
Going t close this PR. Thanks a lot!

AndrewRademacher:
Thank you for pushing these features into the library!

jeremychone:
@AndrewRademacher You're welcome, and thanks for the note and for your work. It helped a lot.

I'm looking at flattening the message content, but this is more involved and intrusive. I need to try it out in a branch and see how it fits. I think it will be a better approach, but it changes quite a bit.

PR #91, opened 2025-10-15
Fixed streaming tool calls for openai models

bartCarroll:
**Problem:**

- OpenAI streams tool call arguments in small chunks (character by character)
- Original code attempted to parse JSON after each chunk
- Parsing incomplete JSON like {" or command failed and stripped quotes
- Result: malformed JSON like {"command:value,timeout_secs":60} instead of {"command":"value","timeout_secs":60}

**Solution:**

- Store argument chunks as raw strings without parsing
- Accumulate all chunks as a concatenated string
- Let downstream consumers parse the complete JSON at stream end

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 `.` minor | `-` Fix | `+` Addition | `^` improvement | `!` Change | `*` Refactor
 
+
+## 2025-10-12 - [v0.4.2](https://github.com/jeremychone/rust-genai/compare/v0.4.1...v0.4.2)
+
+- `.` test - make the common_test_chat_stop_sequences_ok more resilient
+- `^` Anthropic - preserve the content order as it appears in the JSON array (#89)
+- `.` Gemini - when no response, return error with finishReason and usageMetadata
+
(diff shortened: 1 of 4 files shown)


jeremychone:
@AndrewRademacher @bartCarroll 

This looks like a good and needed fix. 

Let me look at it, but based on a quick first scan, it looks good. I will look at it more.

jeremychone:
@bartCarroll I added this "fix" #4123ee2db51a4dcb461e85cf22ac5733ffcae958 to attempt to parse the `fn_arguments` (if it is a string) for the `Stream:End` `captured_tool_calls`.

I think this is the expected behavior.

I also added the first test for `test_chat_stream_tool_capture_ok` (only for OpenAI for now).

I will deal with the `...[index..]` later.

Please let me know if my fix caused any issues.

bartCarroll:
Thanks @jeremychone will test in the morning and report back!

jeremychone:
@bartCarroll Awesome, thanks for checking this out. Happy coding!

jeremychone:
@bartCarroll 0.4.3 released with your change.

PR #96, opened 2025-10-29
Add support for audio inference with OpenAI audio models with the chat completion API

vagmi:
These models offer pretty fast inference for audio and are significantly cheaper than the OpenAI realtime models.


jeremychone:
@vagmi So, this looks very good, but I am not a fan of having a large binary file (i.e., >80KB) in the repo (the .wav file).

If it works with .mp3 (which it might not), that would be great. We might be able to get one around 50KB, but WAV files are just too large.

Otherwise, let's remove it and have the test indicate that it requires this file (just a print, no error if the file is missing). Then, we can add a link to download it in the test with instructions on where to put it. 

So, if you could remove the .wav file, hopefully the squash will do the right thing and not add it to the history.

jeremychone:
@vagmi I merged the first commit, and then revised the second one related to the test so that the `.wav` file is not checked in. The developer will need to have that file eventually, I might use a `download_if_needed`. The test just skips itself if the `.wav` file is not present.

So, this PR can be considered merged, even if the second commit was changed a little to avoid having the 100kb+ test data file in the git history.

Thanks a lot for the support.

PR #100, opened 2025-11-15
. fix - ollama - all_model_names to honor custom endpoints

AlexandreCassagne:
This PR fixes Ollama's list models feature, which is currently hardcoded to use localhost.

### DIFF
--- a/src/adapter/adapters/ollama/adapter_impl.rs
+++ b/src/adapter/adapters/ollama/adapter_impl.rs
@@ -1,5 +1,6 @@
 //! API DOC: https://github.com/ollama/ollama/blob/main/docs/openai.md
 
+use crate::adapter::adapters::support::get_api_key;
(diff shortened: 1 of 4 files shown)


src/adapter/adapters/ollama/adapter_impl.rs
```diff
@@ -32,31 +33,14 @@ impl Adapter for OllamaAdapter {
 	/// TODO: This will use the default endpoint.
 	///       Later, we might add another function with an endpoint, so the user can provide a custom endpoint.
```
AlexandreCassagne:
```suggestion
```

jeremychone:
@AlexandreCassagne Ok, this seems like a good fix.

Now, if you change the Trait API, please open an issue or start a discussion so we can talk it through beforehand.

PR #101, opened 2025-11-23
feat: gemini thought signature

ii64:
Gemini Thought Signature is mandatory for Gemini 3.0

### DIFF
--- a/examples/c10-tooluse-streaming.rs
+++ b/examples/c10-tooluse-streaming.rs
@@ -3,11 +3,13 @@ use genai::Client;
 use genai::chat::printer::{PrintChatStreamOptions, print_chat_stream};
 use genai::chat::{ChatMessage, ChatOptions, ChatRequest, Tool, ToolResponse};
 use genai::chat::{ChatStreamEvent, ToolCall};
+use genai::resolver::AuthData;
 use serde_json::json;
 use tracing_subscriber::EnvFilter;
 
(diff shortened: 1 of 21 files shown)


jeremychone:
Ho make sense. Interesting, I guess this is their alternative to call_id.

jeremychone:
@ii64 The thing is that I have not looked too much into this `thought signatures` thing. It feels related, but only loosely related to `call_id`. But this is just a quick guess.

So I just want to make sure we do not corner `genai` too much with this "Gemini-specific" thing, but on the other hand, it might be very important to support. So `genai` should not be the lowest common denominator.

So maybe it is fine to add it the way you did, and then see later if we want to build consolidation bridges.

Side note: Images should be supported, I think they are for OpenAI. Not on the generated side though. There should be another ticket for this.

Anyway, this is good work, let's think about it. Feel free to add your perspective. I am not against this PR, just want to make sure we do the right thing.

jeremychone:
@ii64 I took some time to look at Thought’s signature, and I think you are right, we need to support it in `genai` because, for Gemini 3 Pro, it is the only way to have function calling (as you mentioned).

So let me review this PR, and I will probably merge it.

Thanks a lot for this great work.

jeremychone:
@ii64 ok, Merged, I am going to bum the release to `0.5.0` so that we do not compile break any code (in case dev create `ToolCall` from props.

PR #103, opened 2025-12-04
Capture response body in ResponseFailedNotJson error

### LINKED ISSUE #102: ResponseFailedNotJson discards response body
When a non-JSON response is received, ResponseFailedNotJson only captures the content-type but discards the body. This makes debugging difficult when APIs return HTML error pages with useful information.

ResponseFailedStatus already captures the body correctly - the same pattern should apply to ResponseFailedNotJson.

https://github.com/jeremychone/rust-genai/blob/2f315aa186e212e2edbd23b2804fdd6205082309/src/webc/error.rs#L10-L19
https://github.com/jeremychone/rust-genai/blob/2f315aa186e212e2edbd23b2804fdd6205082309/src/webc/web_client.rs#L110-L115

Was this intentional to prevent large HTML responses from cluttering logs?

Himmelschmidt:
Fixes #102

Captures the response body in ResponseFailedNotJson, matching the pattern used by ResponseFailedStatus.

### DIFF
--- a/src/chat/mod.rs
+++ b/src/chat/mod.rs
@@ -3,26 +3,26 @@
 
 // region:    --- Modules
 
+mod binary;
 mod chat_message;
 mod chat_options;
 mod chat_req_response_format;
 mod chat_request;
 mod chat_response;
 mod chat_stream;
-mod binary;
 mod content_part;
 mod message_content;
 mod tool;
 mod usage;
 
 // -- Flatten
+pub use binary::*;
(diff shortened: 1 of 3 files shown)


jeremychone:
@Himmelschmidt Makes sense. I originally did not do it because I do not like to capture potentially very big data, but I understand the overall value will be better like this.

Later, maybe we can trim at 1KB or 4KB.

Anyway, good PR, thanks!

PR #108, opened 2025-12-28
Fix Ollama reasoning streaming

mengdehong:
## Summary
  - Handle Ollama OpenAI-compatible streaming where reasoning is emitted in
  `delta.reasoning`
  - Avoid emitting empty content chunks so reasoning is surfaced
  - Add streaming tests for qwen3 and deepseek-r1

  ## Tests
  - cargo test --test tests_p_ollama_reasoning
  test_chat_stream_reasoning_chunk_ok -- --nocapture
  - cargo test --test tests_p_ollama_reasoning
  test_chat_stream_non_empty_chunk_deepseek_ok -- --nocapture

### DIFF
--- a/src/adapter/adapters/openai/streamer.rs
+++ b/src/adapter/adapters/openai/streamer.rs
@@ -253,7 +253,9 @@ impl futures::Stream for OpenAIStreamer {
 
 								// Return the Event
 								return Poll::Ready(Some(Ok(InterStreamEvent::Chunk(content))));
-							} else if let Some(reasoning_content) = reasoning_content {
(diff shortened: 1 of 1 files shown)


jeremychone:
@mengdehong Thanks, i will review it as well, and will commit it. 

It's good to use AI, but always have to tripple check.

jeremychone:
@mengdehong Thanks, the code is clean, and everything looks good. 

I am going to make another change in the streaming code to avoid accessing by index `[]`, since it can crash. But this was there before.

PR #115, opened 2026-01-12
inject skip_thought_signature_validator into Gemini 3 tool call thoughtSignature if None

vlmutolo:
Gemini 3 models require that all tool calls are accompanied by a valid thought signature. They will error out and return a 400 response if none is found. They provide a (discouraged) escape hatch for cases where there cannot be a valid thought signature, such as if tool calls are being copied from a conversation with a different model or if a user wants to pre-seed a conversation including tool calls.

This PR finds cases where a Gemini 3 model is about to be called with tool calls missing thought signatures and inserts the special string `skip_thought_signature_validator` in place of the thought signature.

Future work / not implemented in this PR:
- Maybe the same logic should be applied if there are tool thought signatures from a different model. I imagine this would involve extending the ThoughtSignature variant with a field carrying information about what model generated it.

[Relevant Gemini documentation](https://ai.google.dev/gemini-api/docs/thought-signatures#faqs)


vlmutolo:
It seems like both of those tests would have Gemini generating the actual tool calls. You only hit the error if the tool call request is injected manually instead of being generated by the model.

jeremychone:
@vlmutolo Thanks for the clarification and great PR. Merged.

PR #116, opened 2026-01-12
Ensure extra headers are applied in exec_chat and exec_chat_stream

BinaryMuse:
While working on a proxy for LLM requests, I found that `ChatOptions::extra_headers` wasn't being applied. Upon investigation, only the OpenAI provider was applying them.

This PR updates the codebase so these are applied unconditionally in `Client::exec_chat` and `Client::exec_chat_stream`.

It was unclear to me if this should go before or after the overrides from `AuthData::RequestOverride`.


jeremychone:
@BinaryMuse Good catch, and good PR. I also think the way you did it before AuthData is the most intuitive.

PR #117, opened 2026-01-13
fix: check HTTP status in WebStream before processing byte stream

BinaryMuse:
If a streaming request results in a non-streaming error response, `WebStream` silently ignores it. This PR updates `WebStream` to check `response.status()` before creating the byte stream, and reads the body of the request to create the error message.

It also adds an `error` field to `Error::WebStream` so the original data can be pulled out.

Happy to fix this up with any suggestions.


jeremychone:
@BinaryMuse Thanks a lot for fixing this issue, and good PR quality.

PR #121, opened 2026-01-19
Include thoughts and capture thoughts are reasoning content

vagmi:
Gemini requires that we use the `includeThoughts` flag in the generation config and capture the thoughts text if the `thought` boolean is set in the response.


ropoctl:
> @ropoctl So before I merge your other PR (which might be a little later since it has an API change), this PR is a good one to merge, right?

i would probably rebase my pr to be on this one after you merge this. i left it as "opt-in" to request thinking, since that is how the gemini api does it- you can set thinking budget without opting into showing thinking. but this this library is careful to extract the reasoning out and put it into its own field in the response. gemini has "opt-in" for this, because thought summaries are basically the same type as text, just with thought:true, so if your code isnt checking thought:true, you will get the thoughts prefixing your response text.

PR #123, opened 2026-01-20
Add include thought summary api option for gemini

ropoctl:


jeremychone:
I am going to close this one as discussed in #121

Feel free to do another PR on top of #121. But if the API changes, let's think carefully about how it should behave and what the default should be (opt-in vs. opt-out)

PR #126, opened 2026-01-22
feat: Add `ModelSpec` to allow for additional model details when making API calls

BinaryMuse:
## Problem

While `ClientConfig` allows for overwriting service target details with the `with_service_target_resolver` family of methods, the resolver function is passed a `ServiceTarget` struct that is inferred based on the string model name. This makes it very difficult to do things like use custom providers or specify a dynamic endpoint for a model. I'm currently formatting my model selection as a "provider::model::endpoint" string, which gets stored as `model_name` in the `ModelIden` and is accessible from the `ServiceTarget` — but this gets quite messy and depends on string manipulation to extract the parts.

## Solution

I explored several options to resolve this, but the most promising one was to use an enum to represent the various model specifications, with a trait implementation to ensure that passing in a string still works for backwards compatibility:

```rust
pub enum ModelSpec {
    Name(String),          // Full inference
    Iden(ModelIden),       // Skip adapter inference
    Target(ServiceTarget), // Full control, skip all resolution
}

impl From<&str> for ModelSpec { ... }
impl From<ModelIden> for ModelSpec { ... }
impl From<ServiceTarget> for ModelSpec { ... }
```

Then change the signatures:

```rust
pub async fn exec_chat_stream(
    &self,
    model: impl Into<ModelSpec>,  // Backward compatible!
    chat_req: ChatRequest,
    options: Option<&ChatOptions>,
) -> Result<ChatStreamResponse>
```

With this, I can pass any struct that implements `Into<ModelSpec>`:

```rust
pub enum ModelSelection {
    Custom { model: String, uri: Option<String> },
    Claude { model: String },
    OpenAI { model: String, uri: Option<String> },
    Ollama { model: String, uri: Option<String> },
}

impl From<&ModelSelection> for ModelSpec {
    fn from(model: &ModelSelection) -> Self {
        match model {
            ModelSelection::Custom { model, uri } => {
                ModelSpec::Target(ServiceTarget {
                    model: ModelIden::new(AdapterKind::Anthropic, model),
                    endpoint: Endpoint::from_owned(
                        uri.clone().unwrap_or("http://localhost:4000/...".to_string()),
                    ),
                    auth: AuthData::Key("".to_string()),
                })
            },
            ModelSelection::Claude { model } => ModelSpec::Name(model.clone()),
            ModelSelection::OpenAI { model, .. } => ModelSpec::Name(model.clone()),
            ModelSelection::Ollama { model, uri } => {
                ModelSpec::Target(ServiceTarget {
                    model: ModelIden::new(AdapterKind::Ollama, model),
                    endpoint: Endpoint::from_owned(
                        uri.clone().unwrap_or("http://localhost:11434".to_string()),
                    ),
                    auth: AuthData::Key("".to_string()),
                }),
            }
        }
    }
}

// elsewhere

let model = ModelSelection::Ollama {
    model: "gpt-oss:2b".to_string(),
    uri: Some("http://localhost:11434".to_string()),
};

let stream = client.exec_chat_stream(&model, chat_request, Some(&chat_options)).await?;
```

---

As always, open to any feedback.

### DIFF
--- a/src/client/client_impl.rs
+++ b/src/client/client_impl.rs
@@ -47,7 +47,7 @@ impl Client {
 	/// Accepts any type that implements `Into<ModelSpec>`:
 	/// - `&str` or `String`: Model name with full inference
 	/// - `ModelIden`: Explicit adapter, resolves auth/endpoint
-	/// - `ServiceTarget`: Returns directly, bypasses resolution
+	/// - `ServiceTarget`: Uses directly, bypasses model mapping and auth resolution
 	pub async fn resolve_service_target(&self, model: impl Into<ModelSpec>) -> Result<ServiceTarget> {
 		self.config().resolve_model_spec(model.into()).await
 	}
@@ -57,7 +57,7 @@ impl Client {
 	/// Accepts any type that implements `Into<ModelSpec>`:
 	/// - `&str` or `String`: Model name with full inference
 	/// - `ModelIden`: Explicit adapter, resolves auth/endpoint
-	/// - `ServiceTarget`: Uses directly, bypasses resolution
+	/// - `ServiceTarget`: Uses directly, bypasses model mapping and auth resolution
 	pub async fn exec_chat(
 		&self,
 		model: impl Into<ModelSpec>,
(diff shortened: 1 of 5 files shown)


jeremychone:
@BinaryMuse Hey, I didn't forget about this one, it's just that we got a few more API change PRs on the web search tools, so I'm planning to include all of those in `0.6.0`

jeremychone:
@BinaryMuse  I merged it into main, under `0.6.0-alpha.1-WIP`

Btw, big thank you. The more I looked at this API, the more I like it. You really solved it in an elegant, flexible, and backward-compatible way. I would not have come up with something better.

jeremychone:
@BinaryMuse ok, so I made some minor changes and updated a few things for ModelIden and related items. 

See the new example at: [c06-model-spec](https://github.com/jeremychone/rust-genai/blob/main/examples/c06-model-spec.rs)

Here is the commit for the API changes to ModelSpec: bdec7a5



I think we are in a very good place with this one. Thanks again.

PR #129, opened 2026-01-28
Capture Anthropic cache tokens in streaming usage

tcdent:
Anthropic's prompt caching returns cache_creation_input_tokens and cache_read_input_tokens alongside input_tokens. These were not being captured during streaming, resulting in usage showing only uncached input tokens.

This adds capture of:
- cache_creation_input_tokens -> prompt_tokens_details.cache_creation_tokens
- cache_read_input_tokens -> prompt_tokens_details.cached_tokens

This enables accurate context window tracking when using prompt caching.


jeremychone:
@tcdent Thanks. There are quite a few other PRs on Anthropic right now, which I'm still not sure how to handle. So yes, waiting a little would be good. But this looks like a good one.

PR #130, opened 2026-01-28
fix(anthropic): prompt caching fixes

holovskyi:
## Summary

Two bug fixes for Anthropic prompt caching:

1. **Prompt caching not applied for single system message** (`adapter_impl.rs`)
   - The condition `last_cache_idx > 0` failed when there was only one system message with `cache_control` because index `0 > 0` is false
   - Changed to `>= 0` to correctly distinguish between "no cache_control found" (-1) and "cache_control found at index 0+"

2. **Streaming adapter ignores cache tokens** (`streamer.rs`)
   - Previously only captured `input_tokens` and `output_tokens`, ignoring `cache_creation_input_tokens` and `cache_read_input_tokens`
   - Now captures cache tokens from `message_start` event and adds them to `prompt_tokens` (matching `into_usage()` behavior)
   - Populates `prompt_tokens_details` with cache info

## Test plan

- [x] `cargo check` passes
- [x] Unit tests pass
- [ ] Manual testing with Anthropic API and prompt caching enabled

### DIFF
--- a/examples/c18-cache-ttl.rs
+++ b/examples/c18-cache-ttl.rs
@@ -1,0 +1,179 @@
+//! Example demonstrating prompt caching with mixed TTLs (1h + 5m).
+//!
+//! This example shows how to:
+//! 1. Use `CacheControl::Ephemeral1h` and `CacheControl::Ephemeral5m` on system messages
+//! 2. Verify cache creation on the first request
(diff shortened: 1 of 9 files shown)


jeremychone:
released in `0.6.0-alpha.1`

PR #133, opened 2026-01-28
feat(anthropic): add web search and web fetch tool support

holovskyi:
## Summary
- Add `WebSearchConfig` and `WebFetchConfig` for Anthropic's server-side web tools
- Parse web search results (citations, search result items) and web fetch content from responses
- Track `web_search_requests` and `web_fetch_requests` in `CompletionTokensDetails`
- Add `ServerToolError` type for tool-level error handling
- Include examples `c16-web-search` and `c17-web-fetch`

### DIFF
--- a/examples/c16-web-search.rs
+++ b/examples/c16-web-search.rs
@@ -73,7 +73,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
 	let location = UserLocation::default()
 		.with_city("Tokyo")
-		.with_country("Japan")
+		.with_country("JP") // ISO 2-letter country code
 		.with_timezone("Asia/Tokyo");
 
 	let config = WebSearchConfig::default()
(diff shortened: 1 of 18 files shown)


jeremychone:
Note to self: 

right now, in the gemini adapter, a little "hacky" probably: 

```rust
				// -- if it is a builtin tool
				if matches!(
					req_tool.name.as_str(),
					"googleSearch" | "googleSearchRetrieval" | "codeExecution" | "urlContext"
				) {
					tools.push(json!({req_tool.name: req_tool.config}));
				}
```

jeremychone:
Here is a related [commit for Gemini web search](https://github.com/jeremychone/rust-genai/commit/0b2366e1393689bbd3e3af9e5c0b05af32f7ee67)

Note: I am not saying we should have a typed version of it, but this is just for reference.

jeremychone:
@holovskyi FYI, I’m doing some refactoring of the tool, strongly inspired by what you did here.

I wanted a model that works for multiple providers. It’ll be part of 0.6.0-alpha

The trick is that Google Gemini web search is a little different, in terms of intent, than Anthropic, but I think normalization can still be good.

I’m trying to find a good balance between supporting custom and normalized.

Your code got this work started, and I’m using it a lot to design the new scheme.

PR #134, opened 2026-01-28
fix(anthropic): prevent custom tool hijacking and stale streamer state

holovskyi:
## Summary
- Distinguish native `web_search`/`web_fetch` server tools from user-defined function tools sharing the same name by checking for absent schema
- Handle unknown content block types in the streamer to avoid stale `in_progress_block` state corrupting subsequent delta processing

Depends on #133


jeremychone:
Same as the comments in #133 about being Anthropic-specific and "variant-prolific."

Issue #138, opened 2026-01-30
bug - deepseek-reasoner bug


lazyclaw-labs:
### Bug description & expected

This the error message I got:

Provider error: Provider error: GenAI error: Web call failed for model 'deepseek-reasoner (adapter: DeepSeek)'. Cause: Request failed with status code '400 Bad Request'. Response body: {"error":{"message":"Missing `reasoning_content` field in the assistant message at message index 2. For more information, please refer to https://api-docs.deepseek.com/guides/thinking_mode#tool-calls","type":"invalid_request_error","param":null,"code":"invalid_request_error"}}

After investigatem here is the conclusion from claude code:
  Root cause:

  1. When DeepSeek's reasoner model responds, it includes reasoning_content in the assistant message
  2. The genai library captures reasoning_content in ChatResponse.reasoning_content (line 17 of chat_response.rs)
  3. However, ChatMessage doesn't have a field for reasoning_content
  4. When building the next request, the OpenAI adapter builds assistant messages like this (adapter_impl.rs:489):
  let content = texts.join("\n\n");
  let mut message = json!({"role": "assistant", "content": content});
  4. It never adds reasoning_content to the message JSON

  The empty assistant message you see:

  When the LLM returns only tool calls (no text content), the code produces:
  {"role": "assistant", "content": "", "tool_calls": [...]}

  This is missing the reasoning_content field that DeepSeek requires for the reasoner model.

  This is a genai library issue. The fix needs to:
  1. Store reasoning_content in ChatMessage (new field or as a special ContentPart)
  2. When building OpenAI-compatible requests for assistant messages, include reasoning_content in the JSON if present

  You should file this as an issue/feature request with the genai crate owner. The fix involves modifying:
  - src/chat/chat_message.rs - Add field for reasoning_content
  - src/adapter/adapters/openai/adapter_impl.rs - Include reasoning_content when building assistant message JSON for the request

### Adapter

deepseek

### Model

deepseek-reasoner

PR #147, opened 2026-02-21
fix: reasoning content roundtrip for deepseek/kimi models

Zer0dot:
I had Claude look through this, fix seems straightforward enough, just add a reasoning_content optional parameter to the message and allow sending it back.

This should be seen as a patch while it remains unclear what exactly should be done with these models that now break the OpenAI adapter API.

As an example for how an integrator would use this:

When the LLM responds with tool calls, you send the assistant message back in the next request. Kimi/DeepSeek expect reasoning_content to be  included on that assistant message, or they error out.                                                                                     
```                                                                                                                                             
  let reasoning_content = response.reasoning_content.clone();
  let tool_calls = response.into_tool_calls(); 

  // Build the assistant message for the next LLM call
  let assistant_msg = ChatMessage::from(tool_calls).with_reasoning_content(reasoning_content);
  messages.push(assistant_msg); 
 ```                                                                                                                              
Without this, the provider (e.g. kimi k2.5 thinking) rejects the request because the reasoning chain is broken.

Basically, you need to clone `response.reasoning_content` before consuming the response, and send it back in tool-use loops. Everything else is optional or automatic.

Related issue: https://github.com/jeremychone/rust-genai/issues/138

### DIFF
--- a/src/adapter/adapters/openai/adapter_shared.rs
+++ b/src/adapter/adapters/openai/adapter_shared.rs
@@ -303,7 +303,7 @@ impl OpenAIAdapter {
 
 				// Assistant - For now support Text and ToolCalls
 				ChatRole::Assistant => {
-					// -- If we have only text, then, we jjust returned the joined_texts
+					let reasoning = msg.reasoning_content; // move out before iterating content
 					let mut texts: Vec<String> = Vec::new();
 					let mut tool_calls: Vec<Value> = Vec::new();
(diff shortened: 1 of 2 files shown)


Zer0dot:
I think that makes more sense semantically!

I'll give it a try. It's going to add a lot more match arm cases but it makes more sense to me rather than adding an esoteric top-level field.

PR #154, opened 2026-02-22
feat: add reasoning_content content part

Zer0dot:
Following up from #138 and PR #147.

This pull request refactors reasoning_content into a new `content_part` which, in the openai adapter, is hoisted as a top-level field, similar to Gemini's `thought_signature` field.

This is needed for kimi reasoning models and deepseek as far as I know. The usage is still the same, integrators have access to a  `.with_reasoning_content()`  builder that attaches reasoning content to the `ChatMessage` as a content part.

### DIFF
--- a/src/adapter/adapters/anthropic/adapter_impl.rs
+++ b/src/adapter/adapters/anthropic/adapter_impl.rs
@@ -565,6 +565,7 @@ impl AnthropicAdapter {
 									}));
 								}
 								ContentPart::ThoughtSignature(_) => {}
+								ContentPart::ReasoningContent(_) => {}
 								// Custom are ignored for this logic
 								ContentPart::Custom(_) => {}
 							}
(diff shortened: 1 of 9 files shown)


Zer0dot:
Sorry for the delay.

Confirming that this does work as intended in my dogfooding with streaming output as well.

Zer0dot:
Good point-- I named it reasoningContent cause that's what the kimi and deepseek apis return. Let me know if you think that's reasonable!

E.g. https://api-docs.deepseek.com/guides/reasoning_model#api-example

and https://platform.moonshot.ai/docs/guide/use-kimi-k2-thinking-model#accessing-the-reasoning-content

jeremychone:
Okay, I am merging this like this.

The design is that it allows the caller to pass that reasoning content to a provider/model that supports it. It's okay that the user has to add this. Adding too much logic or too many assumptions has its pitfalls.


Thanks for the PR, and let me know how it goes. I will do a beta release later this week. Ping me here if I forget.

PR #162, opened 2026-03-09
feat(gemini): simplify JSON Schemas for API compatibility

chaizhenhua:
## Problem

Tool schemas generated by `schemars` (or similar libraries) contain standard JSON Schema features that Gemini API rejects with **400 Bad Request**:

- `$ref` / `$defs` — internal references Gemini doesn't resolve
- `allOf` / `anyOf` / `oneOf` — composite type keywords (e.g. schemars `Option<T>` generates `anyOf: [{type: T}, {type: null}]`)
- `"type": ["string", "null"]` — array-type nullable syntax
- `additionalProperties` — already handled by existing `x_walk`, now consolidated

## Solution

New `simplify_schema()` function in `src/adapter/adapters/gemini/schema.rs` that transforms schemas in-place before sending to the Gemini API. Applied to both **tool parameter schemas** and **response format schemas**.

### Transformations (order matters):

1. **Resolve `$ref`** — inline definitions from `$defs`/`definitions` with cycle detection
2. **Remove `$defs`/`definitions`** — after all refs are resolved
3. **Flatten composites** — single-element `allOf`/`anyOf`/`oneOf` merged into parent; nullable variants (schemars `Option<T>` pattern) simplified
4. **Normalize nullable types** — `"type": ["string", "null"]` → `"type": "string"`
5. **Remove `additionalProperties`** — consolidates existing `x_walk` logic
6. **Recurse** into `properties`, `items`, `prefixItems`, remaining composite entries

## Changes

- **New**: `src/adapter/adapters/gemini/schema.rs` — schema simplifier with 15 unit tests
- **Modified**: `src/adapter/adapters/gemini/adapter_impl.rs` — apply `simplify_schema()` to tool params and response format
- **Modified**: `src/adapter/adapters/gemini/mod.rs` — add module declaration

## Tests

All 42 lib tests pass (15 new + 27 existing, zero regressions).

### DIFF
--- a/src/adapter/adapters/gemini/adapter_impl.rs
+++ b/src/adapter/adapters/gemini/adapter_impl.rs
@@ -208,17 +208,9 @@ impl Adapter for GeminiAdapter {
 
 		// -- Response Format
 		if let Some(ChatResponseFormat::JsonSpec(st_json)) = options_set.response_format() {
-			// x_insert
-			//     responseMimeType: "application/json",
-			// responseSchema: {
 			payload.x_insert("/generationConfig/responseMimeType", "application/json")?;
 			let mut schema = st_json.schema.clone();
-			schema.x_walk(|parent_map, name| {
-				if name == "additionalProperties" {
-					parent_map.remove("additionalProperties");
-				}
-				true
-			});
+			super::openapi_schema::to_openapi_schema(&mut schema);
 			payload.x_insert("/generationConfig/responseJsonSchema", schema)?;
 		}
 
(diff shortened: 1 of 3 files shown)


jeremychone:
@fortunto2 Sorry, I merged another good PR on Gemini (#162), and now we have a conflict on this one. 

But this PR is very good, because I got those tool errors once in a while with Gemini, and I was not sure what it was, but I think you nailed it, hopefully it is the same issue.

If you could resolve the conflict, I would be more than happy to check it again. 

BTW, PR #162 is pretty good, I think. It makes the JSON schema more Gemini-resilient, which is awesome. 

Thanks a lot, your PR looks great and is much needed.

PR #163, opened 2026-03-10
fix: Gemini FC protocol — unique call_id, grouped tool responses, tool() constructors

fortunto2:
## Summary

Three fixes discovered while building an autonomous agent loop on top of genai:

- **Unique call_id for Gemini**: Gemini API omits `call_id` in function call responses. The current code reuses `fn_name`, which breaks when the same tool is called multiple times in one turn (e.g., `read_file` called twice). Now synthesizes `"call#name#counter"` identifiers.

- **Merge consecutive tool-response entries**: When sending multiple `ChatMessage::Tool` messages (one per tool response from the agent loop), each creates a separate `{"role":"user"}` entry in the Gemini payload. Gemini requires all `functionResponse` parts in a single turn. Added a post-processing pass that merges consecutive tool-response-only entries.

- **`ChatMessage::tool()` and `tool_responses()` constructors**: The only way to create a Tool message was `From<ToolResponse>`. Added explicit constructors matching the `system()`/`user()`/`assistant()` pattern.

Also refactored thought-signature extraction: replaced nested if-let chains with `take_string`/`take_bool` helpers.

## Test plan

- [x] 4 new unit tests in `gemini::adapter_impl::tests`
- [x] `cargo test -p genai --lib` passes
- [ ] Manual test with multi-tool Gemini agent loop (tested in production with sgr-agent framework)

### DIFF
--- a/src/adapter/adapters/gemini/adapter_impl.rs
+++ b/src/adapter/adapters/gemini/adapter_impl.rs
@@ -402,53 +402,40 @@ impl GeminiAdapter {
 			}
 		};
 
+		let mut tool_call_counter: usize = 0;
 		for mut part in parts {
-			// -- Capture eventual thought signature
-			{
-				if let Some(thought_signature) = part
-					.x_take::<Value>("thoughtSignature")
-					.ok()
-					.and_then(|v| if let Value::String(v) = v { Some(v) } else { None })
-				{
-					content.push(GeminiChatContent::ThoughtSignature(thought_signature));
-				}
-				// Note: sometime the thought is in "thought" (undocumented, but observed in some cases or older models?)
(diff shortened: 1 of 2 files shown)


jeremychone:
@fortunto2 Thanks, this looks like a very clean PR. tests_p_gemini all work, so we're all good.

One little thing I might change in a subsequent commit is the `ChatMessage::tool(_)` vs. `ChatMessage::tool_responses(_)`

So, I implemented `From<Vec<ToolResponse>> for ChatMessage`, so this will be the recommended way.

The `ChatMessage::user/system/assistant` methods are more the exception since they need additional context to create the right message, with the context encoded in the function name.

PR #179, opened 2026-03-24
Feature/OpenAI resp thinking

ropoctl:


ropoctl:
My attempt was for this to be a stacked PR on top of a base of the first PR but github doesnt support diff bases, I'll let you reopen this after the first one is merged

PR #182, opened 2026-03-25
Add yakbak HTTP record/replay integration test infrastructure (openai…

ropoctl:
…_resp)

Adds a lightweight HTTP record/replay test server (yakbak) for deterministic integration testing of the openai_resp adapter without requiring real API keys.

Infrastructure:
- tests/support/yakbak/ — record/replay HTTP proxy server using hyper
- .github/workflows/yakbak-replay-tests.yml — CI workflow

OpenAI Responses API tests:
- Reasoning stream with content + usage assertions
- Tool call stream with function name/args assertions
- Recording scripts for re-capturing cassettes with real API keys

### DIFF
--- a/.github/workflows/yakbak-replay-tests.yml
+++ b/.github/workflows/yakbak-replay-tests.yml
@@ -1,0 +1,33 @@
+name: Yakbak Replay Tests
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+env:
+  CARGO_TERM_COLOR: always
+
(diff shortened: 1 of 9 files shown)


jeremychone:
@ropoctl OK, thanks for this great addition and conservative, focused patch. I merged it.

So, let's go slowly on this one, perhaps starting with `tests_yakback_...` per adapter.

> Note: I like the way you laid out the code in the tests/support/... and the tests_yakback_record with the `record_..` function prefixes.

Issue #181, opened 2026-03-25
Bug: `WebStream` crashes on multi-byte UTF-8 characters at chunk boundaries


ropoctl:
## Summary

`WebStream::poll_next()` in `src/webc/web_stream.rs` calls `String::from_utf8()` on each raw `Bytes` chunk from reqwest's byte stream without buffering incomplete trailing UTF-8 sequences. When a multi-byte character (e.g. CJK/Japanese) straddles a TCP chunk boundary, the stream fatally errors with:

```
incomplete utf-8 byte sequence from index 8191
```

This is not recoverable — the entire stream is lost.

## Reproduction

Use the `openai_resp` adapter with streaming enabled. Send a prompt that elicits a response containing non-Latin characters (Japanese, Chinese, emoji, etc.). With enough tokens, the HTTP chunked transfer encoding will eventually split a multi-byte UTF-8 character across two chunks, causing the error.

In my case, I was processing Japanese mainframe source code (Fujitsu COBOL with kanji comments) and the model's response included those characters. The error occurs intermittently — it depends on where the TCP/HTTP layer decides to split the response body.

The `8191` in the error message suggests an 8KB buffer boundary.

## Root Cause

[`web_stream.rs` line 113](https://github.com/jeremychone/rust-genai/blob/main/src/webc/web_stream.rs#L113):

```rust
let buff_string = match String::from_utf8(bytes.to_vec()) {
    Ok(s) => s,
    Err(e) => return Poll::Ready(Some(Err(Box::new(e) as BoxError))),
};
```

Each `Bytes` chunk from reqwest is decoded as a complete, self-contained UTF-8 string. If a 3-byte UTF-8 character (e.g. `日` = `E6 97 A5`) has its first byte in chunk N and remaining bytes in chunk N+1, `from_utf8` fails on chunk N.

## Call Stack

```
WebStream::poll_next()         — from_utf8() fails on partial trailing bytes
  ↑ polled by
EventSourceStream::poll_next() — line 50, passes Err through (line 80-81)
  ↑ polled by
OpenAIRespStreamer::poll_next() — line 104, wraps into Error::WebStream (line 252-258)
```

## Affected Adapters

Any adapter that uses `WebStream` for SSE parsing. Currently that includes:
- `openai_resp` (OpenAI Responses API) — uses `EventSourceStream` → `WebStream`

The older `openai` chat adapter uses `reqwest-eventsource`/`eventsource-stream` which handles this correctly, so it is **not** affected.

[...]

PR #187, opened 2026-03-29
^ fix: buffer incomplete UTF-8 sequences across stream chunks

ropoctl:
https://github.com/jeremychone/rust-genai/issues/181 WebStream::poll_next() previously called String::from_utf8() on each raw Bytes chunk, which fails when a multi-byte UTF-8 character straddles a chunk boundary (e.g. Japanese 日 split across 8KB TCP chunks).

Fix: add utf8_carry buffer that accumulates incomplete trailing bytes. Uses std::str::from_utf8() to find valid_up_to boundary, converts the valid portion, and carries incomplete trailing bytes to the next chunk. Actual invalid UTF-8 (not just incomplete) still produces a fatal error.

Also changes yakbak replay to use HTTP/1.1 chunked transfer encoding (8192-byte chunks via StreamBody) to realistically reproduce the bug. Includes a yakbak regression test with Japanese text at the 8KB boundary.


jeremychone:
Merged, and released in `0.6.0-beta.13`

PR #199, opened 2026-04-08
fix(openai-streamer): emit ToolCallChunk when tool_calls arrive with finish_reason

chaizhenhua:
## Problem

Providers like ZhiPu (BigModel / GLM-4) send `tool_calls` and `finish_reason` in the same SSE chunk during streaming. The existing streamer code captured the tool call data into `captured_data` but **did not emit an `InterStreamEvent::ToolCallChunk`**, causing agent loops to miss tool calls entirely and treat the response as a text-only NaturalEnd completion.

## Root Cause

In `openai/streamer.rs`, the `finish_reason` branch (line 178) correctly calls `capture_tool_call()` when it finds tool_calls in the same chunk (line 198), but then falls through to `continue` without emitting a stream event. The non-finish_reason branch (line 264) correctly emits `InterStreamEvent::ToolCallChunk`, but it's gated behind `else if` and never reached when finish_reason is present.

## Fix

After capturing tool calls in the finish_reason branch, emit the first captured tool call as `InterStreamEvent::ToolCallChunk` before continuing. This ensures downstream consumers (agent loops, stream collectors) see the tool call event in the stream.

## Testing

Verified with ZhiPu GLM-4-Flash via the `tool_call_live` example:
- Before fix: 1 step, NaturalEnd, empty response (tool call silently lost)
- After fix: 2 steps, tool call executed, correct result returned

## Affected Providers

Any provider that sends `tool_calls` and `finish_reason` in the same SSE chunk. Known: ZhiPu/BigModel (GLM-4 series). Potentially also Ollama (mentioned in existing code comments).

### DIFF
--- a/src/adapter/adapters/openai/streamer.rs
+++ b/src/adapter/adapters/openai/streamer.rs
@@ -208,6 +208,10 @@ impl futures::Stream for OpenAIStreamer {
 							self.captured_data.stop_reason = Some(finish_reason);
 							// NOTE: Some providers (e.g., Ollama) send tool_calls AND finish_reason in the same message.
 							// We need to capture tool_calls here before continuing to the next message.
+							// Capture tool_calls that arrive in the same chunk as finish_reason.
+							// After capturing, emit the first ToolCallChunk so downstream
+							// consumers (e.g. agent loops) see the tool call event.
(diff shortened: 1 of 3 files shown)


jeremychone:
@chaizhenhua released in 0.6.0-beta.16

PR #200, opened 2026-04-09
+ tool - add 'strict' option for OpenAI strict function calling

ropoctl:
Add Tool.strict: Option<bool> so callers can opt into OpenAI's strict function calling mode, which guarantees tool-call arguments conform exactly to the provided JSON Schema.

When strict is true, the adapters:
1. Set "strict": true in the tool JSON
2. Walk the schema to inject "additionalProperties": false on every "type": "object" node (required by the OpenAI API)

This uses the same x_walk pattern already established for structured output (JsonSpec response format).

Supported in both openai (Chat Completions) and openai_resp (Responses API) adapters. Gemini and Anthropic ignore the field.

### DIFF
--- a/src/adapter/adapters/anthropic/adapter_impl.rs
+++ b/src/adapter/adapters/anthropic/adapter_impl.rs
@@ -755,6 +755,7 @@ impl AnthropicAdapter {
 			description,
 			schema,
 			config,
+			..
 		} = tool;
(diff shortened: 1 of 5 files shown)


jeremychone:
@ropoctl released in 0.6.0-beta.16

PR #203, opened 2026-04-10
Add Ollama Cloud adapter

### LINKED ISSUE #202: Add Ollama Cloud adapter
Ollama now offers a cloud API at https://ollama.com that uses the same protocol as local Ollama but requires Bearer token authentication.

This request is to add an `OllamaCloud` adapter that:
- Reuses the existing `OllamaAdapter` for request/response handling
- Injects `Authorization: Bearer <OLLAMA_API_KEY>` on every request
- Defaults to `https://ollama.com/` as base URL
- Supports chat (streaming + non-streaming) and model listing

Env vars: `OLLAMA_API_KEY` (required), `OLLAMA_CLOUD_BASE_URL` (optional override).

RouHim:
Adds an `OllamaCloud` adapter for Ollama's cloud API (https://ollama.com).

Delegates to the existing `OllamaAdapter` and injects Bearer token auth. No changes to the local Ollama adapter.

Includes integration tests, yakbak replay tests, example, docs updates, and CI workflow.

Closes #202

### DIFF
--- a/.github/workflows/yakbak-replay-tests.yml
+++ b/.github/workflows/yakbak-replay-tests.yml
@@ -37,3 +37,6 @@ jobs:
 
       - name: Run GitHub Copilot replay tests
         run: cargo test --test tests_yakbak_github_copilot
+
+      - name: Run Ollama Cloud replay tests
+        run: cargo test --test tests_yakbak_ollama_cloud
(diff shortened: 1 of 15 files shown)


jeremychone:
@RouHim That is cool.

By the way, why not follow the same pattern we use for OpenAI, where OllamaAdapter exposes shared, and then we can reuse this in CloudOllama?

It should be the same protocol for both, right? I have not looked at the Ollama cloud API.

RouHim:
Good idea, I simply never used the local version of ollama. I will try to figure it out, if it's possible.

jeremychone:
@RouHim Cool. Then do another PR so that it's clean and easy to scan.

The nice thing is that now the OllamaAdapter uses the native Ollama protocol, since a PR from a while back. So this should fit nicely in Ollama Cloud.

RouHim:
@jeremychone one other thing (which does not fit into this pr), my previously implemented github copilot pr is missing some functionally. The reason is that the full copilot feature set is just available, if we do have a proper oauth2 authentication. Thus I will send a major rework of the copilot implementation, sorry for the inconvenience.

jeremychone:
Closing as it has been corrected in #204

PR #207, opened 2026-04-14
remote ollama example + `all_model_names` update

wilswer:
Hello!

This PR fixes an issue with the `all_model_names` method on the `Client` struct when using a remote (non-localhost) Ollama instance. Currently, the method uses the default endpoint and therefore only lists local models.

**What's included:**
* **The Fix:** I updated `all_model_names` to take an additional argument, `service_target: Option<ServiceTarget>`, which can hold the custom endpoint.
* **New Example:** I included an example to see it in action, closely related to the `c06-target-resolver.rs` example. (That example works for remote Ollama, but `all_model_names` does not, hence this PR).

I realize adding the argument changes the method signature, so I'm very open to feedback if there is a more elegant way you'd prefer to handle this.

I skipped opening an issue since I thought I could put together the fix myself, but I'm happy to discuss if this needs tweaking.
Thanks for your time!

### DIFF
--- a/examples/c05-model-names.rs
+++ b/examples/c05-model-names.rs
@@ -25,7 +25,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
 	for &kind in KINDS {
 		println!("\n--- Models for {kind}");
-		let models = client.all_model_names(kind).await?;
+		let models = client.all_model_names(kind, None).await?;
 		println!("{models:?}");
 	}
 
--- a/examples/c99-remote-ollama.rs
+++ b/examples/c99-remote-ollama.rs
@@ -1,0 +1,75 @@
+//! This example demonstrates how to use a custom ServiceTargetResolver to
+//! point the genai Client to a remote (non-localhost) Ollama instance.
+
+use genai::adapter::AdapterKind;
+use genai::chat::{ChatMessage, ChatOptions, ChatRequest};
+use genai::resolver::{Endpoint, ServiceTargetResolver};
+use genai::{Client, ModelIden, ServiceTarget};
+use tracing_subscriber::EnvFilter;
+
+// Use a model you have pulled on your remote Ollama server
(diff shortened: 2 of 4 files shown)


jeremychone:
@wilswer YYes, I understand the current limitation and wanted to fix it. 

The trick is that ServiceTarget is `.model`, which seems to be the wrong Struct to use for that purpose. 

I might introduce a type `ProviderCredential` (not sure about the name), with just Endpoint and AuthData. 

Let me think about this one.

jeremychone:
@wilswer I did not forget about this one. I will fix it. I will add a new type with a convenient Into<...>. I just need to think about it.

wilswer:
Thanks @jeremychone! And thank you for developing and maintaining `rust-genai`!

jeremychone:
@wilswer Thanks you for the feedback. Fixed in  eaab4858c990615d848be7deed8074706b015c3c

I fixed this with a new `ProviderConfig` type and bunch of convenient Froms and a 

` 


```rust
// -- v0.5.x

let models = client.all_model_names(adapter_kind).await?;

// -- v0.0.6
// Client::all_model_names(adapter_kind, impl Into<ProviderConfig>)`
// ProviderConfig { 	pub endpoint: Option<Endpoint>, pub auth: Option<AuthData>}

// Same as v0.5.x
let models = client.all_model_names(adapter_kind, None).await?;

// Remote Ollama endpoint override
let models = client
	.all_model_names(
		AdapterKind::Ollama,
		Endpoint::from_static("http://remote-host:11434/"),
	)
	.await?;

// Custom openai compatible provider
let models = client
	.all_model_names(
		AdapterKind::OpenAI,
		(
			Endpoint::from_static("http://fair-llm.ai/"),
			AuthData::from_env("FAIR_LLM_API_KEY"),
		),
	)
	.await?;
```

wilswer:
Thanks @jeremychone!

PR #208, opened 2026-04-18
client: ClientBuilder::with_default_adapter_kind for explicit provider intent

ropoctl:
## Summary

Adds `ClientBuilder::with_default_adapter_kind(kind)` to let a caller pin an authoritative default adapter on the client, so `client.exec_chat("some-model", …)` skips the `AdapterKind::from_model` heuristic (which falls back to `Ollama` for unrecognized names) and uses the configured adapter instead. Opt-in; default behavior is unchanged.

## Motivation

A caller configures `AuthResolver` + `ServiceTargetResolver` scoped to, say, `AdapterKind::OpenAI`, and points at a custom `base_url` — MiniMax, Together, OpenRouter, DeepSeek, a self-hosted vLLM, any gateway that speaks the OpenAI Chat Completions wire protocol. Then:

```rust
    .with_auth_resolver(/* key, scoped to OpenAI */)
    .with_service_target_resolver(/* endpoint = "https://api.minimax.io/v1", scoped to OpenAI */)
    .build();

client.exec_chat("MiniMax-M2.7", req, opts).await?;
```

`AdapterKind::from_model("MiniMax-M2.7")` returns `AdapterKind::Ollama` via the fallback. The resolvers are scoped to `AdapterKind::OpenAI`, so both gates miss, both overrides drop, and the request silently routes to the default Ollama endpoint (`localhost:11434`) with no auth. From the caller's perspective, the `base_url` override "didn't work" — hard to debug without reading resolver internals.

Users today can work around it with `openai::MiniMax-M2.7` prefix namespacing, but:
1. It's not discoverable — a fresh caller has no reason to suspect it.
2. The fallback-to-Ollama behavior is a correctness trap, not a hint.
3. When a client is already scoped to one provider at construction time, forcing the user to re-declare the adapter on every `exec_chat` call duplicates intent.

## Change

```rust
    .with_default_adapter_kind(AdapterKind::OpenAI)
    .with_auth_resolver(/* ... */)
    .with_service_target_resolver(/* ... */)
    .build();

client.exec_chat("MiniMax-M2.7", req, opts).await?;  // now routes correctly
```

- `ClientConfig::default_adapter_kind: Option<AdapterKind>` field + `with_default_adapter_kind` setter + getter.
- `ClientBuilder::with_default_adapter_kind(kind)` passthrough.
- `resolve_model_spec` — for the `ModelSpec::Name` branch, if the default is set and the name is not explicitly namespaced, use the default; otherwise fall back to `AdapterKind::from_model` as before.
- `AdapterKind::from_model_namespace` bumped from `fn` to `pub(crate)` so `resolve_model_spec` can consult it before applying the default. Preserves the existing precedence: explicit namespacing (`openai::some-model`) and `ModelSpec::Iden` calls always win over the client's default.

## Back-compat

Default is `None`, so existing behavior is unchanged for every caller that doesn't opt in. The change is strictly additive.

## Tests

All 61 existing library tests pass. Happy to add a dedicated test for the new path if you'd like — point me at the style you prefer (`tests/` integration test vs inline `#[cfg(test)]`).

## Context

I maintain `genai-pyo3` (the Python bindings for `genai`), and this issue surfaced through a downstream user trying to target MiniMax via the OpenAI-compatible adapter. They'd written ~180 lines of aiohttp as a workaround before we traced the root cause here. I'm running the fork (`ropoctl/rust-genai`, branch `feat/default-adapter-kind`, same commit as this PR) in `genai-pyo3 v0.1.10` in the interim. Happy to flip back to the upstream crate as soon as this lands.

### DIFF
--- a/src/adapter/adapter_kind.rs
+++ b/src/adapter/adapter_kind.rs
@@ -256,7 +256,7 @@ impl AdapterKind {
 
 /// Inner api to return
 impl AdapterKind {
-	fn from_model_namespace(model: &str) -> Option<Self> {
+	pub(crate) fn from_model_namespace(model: &str) -> Option<Self> {
 		let (namespace, _) = ModelName::split_as_namespace_and_name(model);
 		let namespace = namespace?;
 
--- a/src/client/builder.rs
+++ b/src/client/builder.rs
@@ -1,3 +1,4 @@
+use crate::adapter::AdapterKind;
 use crate::chat::ChatOptions;
 use crate::resolver::{
 	AuthResolver, IntoAuthResolverFn, IntoModelMapperFn, IntoServiceTargetResolverFn, ModelMapper,
@@ -92,6 +93,19 @@ impl ClientBuilder {
 		let client_config = self.config.get_or_insert_with(ClientConfig::default);
 		let model_mapper = ModelMapper::from_mapper_fn(model_mapper_fn);
 		client_config.model_mapper = Some(model_mapper);
+		self
+	}
+
+	/// Set a default [`AdapterKind`] for resolving bare model names (creates
+	/// `ClientConfig` if absent).
+	///
+	/// See [`ClientConfig::with_default_adapter_kind`] for details. Useful
+	/// when you've already pinned the provider via auth / service-target
+	/// resolvers and want unrecognized model names to route through that
+	/// provider instead of falling back to Ollama.
+	pub fn with_default_adapter_kind(mut self, adapter_kind: AdapterKind) -> Self {
+		let client_config = self.config.get_or_insert_with(ClientConfig::default);
+		client_config.default_adapter_kind = Some(adapter_kind);
 		self
 	}
 }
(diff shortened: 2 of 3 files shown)


jeremychone:
@ropoctl So this is the `fallback adapter`, right? 

So perhaps calling it `fallback_adapter` would be more explicit, no? `default_adapter` feels a little overreaching to me.

ropoctl:
Agreed — and after sitting with your pushback I think `default_` and `fallback_` both undersell what's actually going on, so I'd rather withdraw this one.

The deeper observation is that a `Client` with `AuthResolver` + `ServiceTargetResolver` already pinned to `AdapterKind::X` is *physically* single-provider — both resolvers are gated on `adapter_kind == X`, so any call that routes through a different adapter silently bypasses auth and the custom endpoint. Today the Client pretends to be multi-provider via model-string sniffing, but that illusion breaks the moment configuration enters. A "fallback" knob patches the symptom; storing the adapter on the Client matches the configuration that's already there.

Follow-up PR coming shortly: a single `ClientBuilder::with_adapter_kind(AdapterKind)` that makes the Client type-level bound to one adapter — `exec_chat("some-model", ...)` routes through it directly, no inference, and `ModelSpec::Iden` / namespace-prefixed names with a mismatched adapter become hard errors (matches what the resolvers would've done silently anyway). Zero-config `Client::default()` is unchanged.

Closing this one in favor of that. Will link the new PR here when it's up.

PR #210, opened 2026-04-18
Add GitHub Copilot OAuth device flow adapter

### LINKED ISSUE #209: Add GitHub Copilot OAuth device flow adapter
Add GitHub Copilot OAuth device flow support, refresh the adapter docs/examples, and cover the new stream and replay behavior.

RouHim:
Closes #209

What's done:
- Add GitHub Copilot adapter with OAuth device flow
- update docs and examples for the new auth path
- add tests and replay coverage for Copilot streams


jeremychone:
@RouHim Not sure this should be part of `genai`. It needs a round trip and all. So, the library user should have the correct key and token in the environment or in the AuthData, and they can do the auth ping-pong with the service.

RouHim:
Ok, valid point. Thank you for your feedback.
I'll move the oauth flow out of the lib.

RouHim:
Oof, I think we have to do answer a base architecturally decision here.

Copilot auth is two stages:

1. On first interactive login / auth we've got a one-time long-lived GitHub OAuth token (ghu_…)
2. Session token (non-interactive, recurring, based on the Stage 1 token, Returns a short-lived (~30 min) bearer token

Stage 2 means a static env var or a static AuthData::Key can't really represent "the Copilot token", whatever value we put in there goes stale every 30 minutes.

Sooo, I think we have 3 options:
1) Genai stays "dumb": Caller passes a fresh Stage 2 token per call (re-setting env var or AuthData before each request). Pushes a non-trivial refresh loop and concurrency story onto every Copilot consumer.
2) Add some kind of AuthData::Dynamic(...) Caller passes a clojure. genai invokes it before each request. Caller still owns the ping-pong. Genai gains a new feature that auth value can be lazily computed. Could be re-used for other token-refreshing providers like Bedrock STS or Vertex
3) Only the interactive device flow leaves genai; the copilot stage 1 and stage 2  + 30-min cache stays, the genai copilot adapter handles the trivial stateless exchange internally, same as it'd handle, say, an STS exchange.
The upstream usage OAuth thing is gone, but the protocol mechanics stay where the protocol knowledge lives.

jeremychone:
@RouHim Need more thinking about this. Your comments are giving good context and decision points.

RouHim:
Take your time <3

ropoctl:
I do this pattern with other OAuth providers, but I think this is better left out of scope for this library

jeremychone:
@ropoctl I just need to make sure that one can implement OAuth their own way and use genai, which I think is the case right now. However, we can think about adding something.

ropoctl:
> @ropoctl I just need to make sure that one can implement OAuth their own way and use genai, which I think is the case right now. However, we can think about adding something.

https://github.com/Lazarus-AI/clearwing/blob/88d3a8a41c22ad9c4d8aaf67bbf5489549740454/clearwing/llm/native.py#L129 this is how we handle in one case- its not rust-genai directly, but the python pyo3 wrapper is 99% faithful in api to rust-genai.

jeremychone:
@RouHim I am going to close that one, but thanks for bringing up this topic. 

Let me know if the `genai` library makes it difficult to implement this use case on top of it, as AuthResolver and ServiceTargetResolver should be helpful for those.

RouHim:
Thank you for your time.

PR #212, opened 2026-04-20
client: ClientBuilder::with_adapter_kind — bind a Client to one adapter

ropoctl:
Supersedes #208. After @jeremychone's pushback on the `default_` / `fallback_` naming there, I re-examined what's actually happening and the answer is that "default" undersold it. This PR picks up the same motivation but reframes and sharpens the behavior.

## Observation

A `Client` whose `AuthResolver` and/or `ServiceTargetResolver` are scoped to a specific `AdapterKind` is already *physically* single-provider — both resolvers gate on `adapter_kind == X`:

```rust
.with_auth_resolver(AuthResolver::from_resolver_fn(move |iden: ModelIden| {
    if iden.adapter_kind == AdapterKind::OpenAI { Ok(Some(auth.clone())) } else { Ok(None) }
}))
.with_service_target_resolver(ServiceTargetResolver::from_resolver_fn(move |mut t| {
    if t.model.adapter_kind == AdapterKind::OpenAI { t.endpoint = custom_endpoint.clone(); }
    Ok(t)
}))
```

Any call whose resolved `adapter_kind` differs from what the resolvers expect silently drops both the auth and the custom endpoint. The request then hits an adapter default with no auth. From the caller's perspective this looks like "my base_url override didn't work" — hard to debug without reading resolver internals.

Today the routing decision is `AdapterKind::from_model(&name)` — a chain of hardcoded `starts_with` heuristics (`gpt-*`, `claude-*`, `gemini-*`, `grok-*`, `deepseek-chat`, `command-*`, `embed-*`, `glm-*`, `mimo-*`, `*fireworks*`, `o1/o3/o4`, `chatgpt`, `codex`, `text-embedding`) with Ollama as the catch-all. That heuristic works for the quick-start case (`exec_chat(\"gpt-4\", ...)` just works) but is:

- **Brittle to new model families** — a new OpenAI prefix silently routes to Ollama until a library update.
- **Wrong for proxies and gateways** — MiniMax, OpenRouter, DeepSeek gateways, self-hosted vLLM, Azure deployment names, any proxy serving the OpenAI wire protocol under a non-`gpt-*` name — all fall through to Ollama.
- **The wrong layer to decide** when the caller has already declared the provider via resolvers.

## Change

Let the caller declare the constraint directly:

```rust
    .with_adapter_kind(AdapterKind::OpenAI)
    .with_auth_resolver(/* ... */)
    .with_service_target_resolver(/* endpoint = \"https://proxy.example/v1\" */)
    .build();

client.exec_chat(\"MiniMax-M2.7\", req, opts).await?;  // routes via OpenAI — matches the resolvers
```

Routing behavior when `adapter_kind` is set:

| `ModelSpec` variant | Bound-adapter behavior |
|---|---|
| `Name(name)` | Routes through bound adapter. Bare names route directly; an explicit `::` namespace that resolves to a *different* adapter returns `Error::AdapterKindMismatch` (previously: silent drop). |
| `Iden(model)` | Bound adapter must match the Iden's `adapter_kind`; otherwise returns `Error::AdapterKindMismatch`. |
| `Target(target)` | Unchanged. A fully-resolved target has already opted out of Client-level routing. |

When `adapter_kind` is unset, behavior is **identical to before** — classic multi-provider-via-model-string, Ollama fallback included. Strictly additive.

## Error shape

```rust
Error::AdapterKindMismatch {
    bound: AdapterKind,
    requested: AdapterKind,
    model: String,
}
```

Display includes a pointer to either drop the `::` namespace or build a Client without `with_adapter_kind` for per-call routing.

## Tests

5 new unit tests in `src/client/config.rs`:

- `bound_client_routes_bare_name_through_bound_adapter` — `mini-max-m2.7` (hits no prefix matcher) routes through the bound OpenAI adapter and the custom-endpoint resolver fires.
- `bound_client_rejects_mismatched_namespace` — `anthropic::claude-*` against an OpenAI-bound Client → `AdapterKindMismatch { bound: OpenAI, requested: Anthropic, .. }`.
- `bound_client_accepts_matching_namespace` — redundant `openai::gpt-4` on an OpenAI-bound Client proceeds normally.
- `bound_client_rejects_mismatched_iden` — `ModelSpec::Iden(Gemini, ...)` on an OpenAI-bound Client → mismatch.
- `unbound_client_preserves_inference` — no `with_adapter_kind` set → `gpt-4` infers to OpenAI exactly as today.

All 61 existing library tests still pass (66 total with the new ones). No changes to integration tests.

## Exposed

- `ClientConfig::adapter_kind: Option<AdapterKind>` field
- `ClientConfig::with_adapter_kind(AdapterKind)` chainable setter
- `ClientConfig::adapter_kind()` getter
- `ClientBuilder::with_adapter_kind(AdapterKind)` passthrough
- `Error::AdapterKindMismatch { bound, requested, model }`
- `AdapterKind::from_model_namespace` bumped from `fn` to `pub(crate)` so the Name-path can detect a namespaced name and fill `requested` in the error.

## Back-compat

Default is `None`. Every caller that doesn't call `with_adapter_kind` sees zero behavior change.

## Context

I maintain `genai-pyo3` (the Python bindings). This issue first surfaced through a downstream user targeting MiniMax via the OpenAI-compatible adapter — they'd written ~180 lines of aiohttp as a workaround before we traced it to the silent Ollama fallback. I'm running this change on the `ropoctl/rust-genai` fork (`feat/bind-client-to-adapter-kind`) in `genai-pyo3 v0.1.13` in the interim. Happy to flip back to the upstream crate as soon as this lands.

### DIFF
--- a/src/adapter/adapter_kind.rs
+++ b/src/adapter/adapter_kind.rs
@@ -256,7 +256,7 @@ impl AdapterKind {
 
 /// Inner api to return
 impl AdapterKind {
-	fn from_model_namespace(model: &str) -> Option<Self> {
+	pub(crate) fn from_model_namespace(model: &str) -> Option<Self> {
 		let (namespace, _) = ModelName::split_as_namespace_and_name(model);
 		let namespace = namespace?;
 
--- a/src/client/builder.rs
+++ b/src/client/builder.rs
@@ -1,3 +1,4 @@
+use crate::adapter::AdapterKind;
 use crate::chat::ChatOptions;
 use crate::resolver::{
 	AuthResolver, IntoAuthResolverFn, IntoModelMapperFn, IntoServiceTargetResolverFn, ModelMapper,
@@ -92,6 +93,19 @@ impl ClientBuilder {
 		let client_config = self.config.get_or_insert_with(ClientConfig::default);
 		let model_mapper = ModelMapper::from_mapper_fn(model_mapper_fn);
 		client_config.model_mapper = Some(model_mapper);
+		self
+	}
+
+	/// Bind the Client to a single [`AdapterKind`] (creates `ClientConfig` if absent).
+	///
+	/// See [`ClientConfig::with_adapter_kind`] for semantics. Short version:
+	/// a Client whose `AuthResolver` / `ServiceTargetResolver` are gated on an
+	/// adapter is already physically single-provider; this makes that
+	/// constraint explicit and drives routing directly instead of inferring
+	/// the adapter from the model name on every call.
+	pub fn with_adapter_kind(mut self, adapter_kind: AdapterKind) -> Self {
+		let client_config = self.config.get_or_insert_with(ClientConfig::default);
+		client_config.adapter_kind = Some(adapter_kind);
 		self
 	}
 }
--- a/src/client/client_impl.rs
+++ b/src/client/client_impl.rs
@@ -27,6 +27,19 @@ impl Client {
 		Ok(models)
 	}
 
+	/// Returns the bound [`AdapterKind`] for this Client, if any.
+	///
+	/// Set via [`ClientBuilder::with_adapter_kind`] at construction
+	/// time — `None` for Clients that did not opt into the bound
+	/// shape and still use per-call model-name inference. Useful for
+	/// callers that want to introspect a Client without tracking the
+	/// provider in parallel state.
+	///
+	/// [`ClientBuilder::with_adapter_kind`]: crate::ClientBuilder::with_adapter_kind
+	pub fn adapter_kind(&self) -> Option<AdapterKind> {
+		self.config().adapter_kind()
+	}
+
 	/// Builds a ModelIden by inferring AdapterKind from the model name.
 	pub fn default_model(&self, model_name: &str) -> Result<ModelIden> {
 		// -- First get the default ModelInfo
--- a/src/client/config.rs
+++ b/src/client/config.rs
@@ -14,6 +14,7 @@ pub struct ClientConfig {
 	pub(super) web_config: Option<WebConfig>,
 	pub(super) chat_options: Option<ChatOptions>,
 	pub(super) embed_options: Option<EmbedOptions>,
+	pub(super) adapter_kind: Option<AdapterKind>,
 }
 
 /// Chainable setters related to the ClientConfig.
@@ -60,6 +61,33 @@ impl ClientConfig {
 		self
 	}
 
+	/// Binds this Client to a single [`AdapterKind`].
+	///
+	/// A Client that has configured an [`AuthResolver`] or [`ServiceTargetResolver`]
+	/// scoped to a specific adapter is *already* physically single-provider: both
+	/// resolvers gate on `adapter_kind`, so any call that routes through a
+	/// different adapter silently drops auth and the configured endpoint. Setting
+	/// `adapter_kind` on the Client makes that constraint explicit and drives
+	/// routing directly, bypassing the [`AdapterKind::from_model`] name-sniffing
+	/// heuristic (which falls back to Ollama for unrecognized names).
+	///
+	/// When set:
+	/// - [`ModelSpec::Name`] routes through this adapter. A `::` namespace prefix
+	///   or other embedded adapter reference is a usage error (see
+	///   [`Error::AdapterKindMismatch`]).
+	/// - [`ModelSpec::Iden`] must carry the same adapter; otherwise returns
+	///   [`Error::AdapterKindMismatch`] instead of silently producing a
+	///   misconfigured request.
+	/// - [`ModelSpec::Target`] is passed through unchanged (callers handing in a
+	///   fully-resolved target have already opted out of Client-level routing).
+	///
+	/// Leave unset for the classic multi-provider shape (route per-call by model
+	/// name).
+	pub fn with_adapter_kind(mut self, adapter_kind: AdapterKind) -> Self {
+		self.adapter_kind = Some(adapter_kind);
+		self
+	}
+
 	/// Returns the WebConfig, if set.
 	pub fn web_config(&self) -> Option<&WebConfig> {
 		self.web_config.as_ref()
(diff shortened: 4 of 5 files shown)


jeremychone:
@ropoctl Just to make sure, if we do not do `client::with_adapter_kind`, then, same behavior as before, right?

ropoctl:
Correct

ropoctl:
Force-pushed one amendment: added `Client::adapter_kind()` on `client_impl.rs` so the bound adapter is readable through the public Client surface, not just `ClientConfig` (which is `pub(crate)`). Matching unit test `bound_client_exposes_adapter_kind_via_getter` pinned in `config.rs`.

Catches the case where a caller wants to introspect a built Client without carrying the `AdapterKind` alongside it in parallel state. Otherwise the setter was essentially write-only from outside the crate.

6 new tests now, 67 total (still strictly additive to the original 61).

jeremychone:
@ropoctl Somehow my brain is stuck here. 

You might have said it in the decision matrix, but I am still unsure. 

If I do `with_adapter(OpenAI)`, and then, 

1) Model is `deepseek::...`, what happens? An error? 
2) Model is `opus-4-7`, what happens? 
3) For `ServiceTargetResolver`, I assume it gets `AdapterKind::OpenAI`, but can it return something else?


### Thinking 

I agree that right now the model resolution flow could be clearer, or at least better documented. 

This is why I want to be careful before we introduce a new scheme. 

So, if a user wants to have a single adapter, why not use and document the service target resolver? 

BTW, don't get me wrong, I am not saying this is not needed, I just want to make sure we are not introducing more complexity.

And once we validate this, I'm all good to add what we decided.

ropoctl:
@jeremychone — fair pushback. Direct answers first, then the ergonomics comparison you're asking about, then an offer to scope the PR down.

### Q1/Q2/Q3 — current PR behavior

1. `with_adapter_kind(OpenAI)` + model = `deepseek::foo` → `Error::AdapterKindMismatch { bound: OpenAI, requested: DeepSeek, model: "deepseek::foo" }`. Hard error, not silent.
2. `with_adapter_kind(OpenAI)` + model = `opus-4-7` → routes through OpenAI. No `::` namespace, so the mapper silently coerces. (This is the asymmetry: explicit namespace = strict reject; bare name = silent coerce. Defensible — namespace is explicit user intent and worth catching when it mismatches; bare names are exactly the case the feature exists to fix — but worth flagging in docs.)
3. `ServiceTargetResolver` receives the bound adapter (after the mapper coerces). It can mutate freely, including `service_target.model.adapter_kind`. Library-level escape hatch by design — the closure is user code, so any constraint would be advisory.

### "Why not just use ServiceTargetResolver" — and what's actually needed

The pipeline order matters here:

```
ModelMapper → AuthResolver → default_endpoint → ServiceTargetResolver
```

ServiceTargetResolver runs *last*. By the time it fires, AuthResolver has already run against whatever `AdapterKind::from_model` returned — so an OpenAI-gated AuthResolver misses for `MiniMax-M2.7` (resolves to Ollama via fallback) and the request goes out with no auth before ServiceTargetResolver gets a chance to fix anything.

The thing that actually does the trick today is **ModelMapper**, not ServiceTargetResolver — it's the only resolver that runs *before* auth. Bind everything via three closures coordinating on the same adapter:

### What it looks like today (no `with_adapter_kind`)

```rust
let mapper = ModelMapper::from_mapper_fn(move |iden: ModelIden| {
    Ok(ModelIden::new(AdapterKind::OpenAIResp, iden.model_name))
});

let auth = AuthResolver::from_resolver_fn(move |iden: ModelIden| {
    if iden.adapter_kind == AdapterKind::OpenAIResp {
        Ok(Some(AuthData::from_single(api_key.clone())))
    } else {
        Ok(None)
    }
});

let target = ServiceTargetResolver::from_resolver_fn(
    move |mut t: ServiceTarget| {
        if t.model.adapter_kind == AdapterKind::OpenAIResp {
            t.endpoint = Endpoint::from_owned(base_url.clone());
        }
        Ok(t)
    },
);

    .with_model_mapper(mapper)
    .with_auth_resolver(auth)
    .with_service_target_resolver(target)
    .build();
```

### What it would look like with the PR

```rust
    .with_adapter_kind(AdapterKind::OpenAIResp)
    .with_auth_resolver_fn(move |_| Ok(Some(AuthData::from_single(api_key.clone()))))
    .with_service_target_resolver_fn(move |mut t| {
        t.endpoint = Endpoint::from_owned(base_url.clone());
        Ok(t)
    })
    .build();
```

Three closures collapse to one boolean + two closures, and the `iden.adapter_kind == X` gate boilerplate disappears (the mapper guarantees it).

### Counter-proposal: scope it down

The strictness in Q1 (`Error::AdapterKindMismatch`) was my call, not load-bearing. Happy to drop it: `with_adapter_kind` becomes a literal "always use this adapter, no name inference, no namespace honored." Smaller surface — no new error variant, no asymmetry to document. Internally it's the equivalent of injecting that `ModelMapper` automatically; nothing else.

If that's easier to defend on review, I'll force-push to drop the namespace-rejection path and the `Error::AdapterKindMismatch` variant. The `Client::adapter_kind()` getter and the test coverage would stay. Let me know which version you'd prefer.

Either way: the mapper-based workaround works today, so this PR is convenience, not a blocker for downstream. genai-pyo3 (and clearwing on top) can ship the workaround pattern in the meantime.

jeremychone:
@ropoctl Sorry, I thought I answered yesterday, but I might not have pressed the button. 

First, thank you for the detailed explanation. 

I like it, and 
- I like the fact that ServiceTargetResolver still has the final say. 
- If not used, it is the same behavior as before. 

I need to spend more time documenting the Model Mapper, AuthResolver with Adapter, and ServiceTargetResolver.

So leave it like this, I will add it. 

Then, this week, I will try to release v0.6.0. I need to work on the migration markdown file.

jeremychone:
@ropoctl Thank you for the great work 

> Sorry for the delay. I worked on enhancing my `pro@coder` with workbench data files so that I can triple-check PRs.

PR #215, opened 2026-05-01
add support for bedrock inference

### LINKED ISSUE #88: Support AWS Bedrock models
Supporting AWS Bedrock models would be helpful. Are there any plans to add that?

Would it be difficult add support? In the beginning only with for example Claude as a supported subset?
Thinking from a potential would like to help out perspective.

clayrosenthal:
Using the aws-config and aws-sigv4 crates to support calling bedrock converse API

closes #88

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,20 @@ unsafe_code = "forbid"
 # unused = { level = "allow", priority = -1 } # For exploratory dev.
 # missing_docs = "warn"
 
+[features]
+default = []
+# AWS Bedrock Converse adapter using SigV4 signing + the full AWS credential chain.
+# Pulls in aws-config (env → profile → SSO → IMDS → AssumeRole) and aws-sigv4 (request signing).
+# Heavy: aws-config transitively adds aws-smithy-*, hyper, tower, h2.
+#
+# The simpler Bearer-token Bedrock adapter (BEDROCK_API_KEY) is always available — no feature
+# needed — since it has no extra dependencies.
+bedrock-sigv4 = [
+	"dep:aws-config",
+	"dep:aws-credential-types",
+	"dep:aws-sigv4",
+]
+
 [dependencies]
 # -- Tracing
 tracing = { version = "0.1", features = ["default"] }
@@ -38,6 +52,10 @@ value-ext = "0.1.3"
 strum = { version = "0.28", features = ["derive"] }
 uuid = { version = "1.21.0", features = ["v4"] }
 regex = "1"
+# -- AWS Bedrock (optional, `bedrock-sigv4` feature)
+aws-config = { version = "1", optional = true }
+aws-credential-types = { version = "1", optional = true }
+aws-sigv4 = { version = "1", optional = true }
 
 [dev-dependencies]
 simple-fs = "0.12"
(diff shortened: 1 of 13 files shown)


jeremychone:
@clayrosenthal Wow, that is great. I looked at Bedrock a few weeks back and got confused by all of those different ways. I was not sure if `invoke` was better than `converse`.

So, `converse` is the right way to go? Same payload for all model types?

Now, one thing I was trying to avoid is bringing in all the `aws-...` stuff, as those are really kitchen sink dependencies.

I was planning to just support the `BEDROCK_API_KEY` method at first.

So, ideally, here is what I thought we could have.

1) A simple `BEDROCK_API_KEY` way, without any `aws-...` and http/crc32fast dependencies, just like any others.

2) Then, have one feature `bedrock-sigv4` which would add support for the kitchen sink, as you have today.

This way, 1) would be included by default, as it will add virtually no weight, with no dependencies.

And 2), if users want the full aws creds support.

What do you think?

clayrosenthal:
> So, converse is the right way to go? Same payload for all model types?
Yeah it's much better.

I'm down to try and take a crack at the simple bedrock API key way. And I can leave the feature for the larger deps.

clayrosenthal:
Ok @jeremychone, I've split it to have both a `bedrock-sigv4` feature with full aws chain and deps, `BedrockSigv4`, and then just a no-extra dep adapter `BedrockApi`. Lmk what you think

jeremychone:
@clayrosenthal Cool, I am planning to look at it today or tomorrow.

jeremychone:
@clayrosenthal Merged, it will be in the beta.19 release. Thank you so much, I really wanted to add this provider, so this is perfect.

clayrosenthal:
Great! Happy to help!

jeremychone:
@clayrosenthal By the way, quick question, so the `converse` API is the new one, and it is better to use now (rather than `invoke`). When I went to their site, I got a little confused. Any insight is welcome.

clayrosenthal:
Yeah the `converse` API is the new one. The benefits of it are that each model takes the same set of inputs and the output is accessed the same way. The old invoke API was much more generic and required custom input and output schemas. For example [claude 2](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-text-completion.html) was different from [claude 3](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages-request-response.html) which was different from [nova](https://docs.aws.amazon.com/nova/latest/userguide/using-invoke-api.html). It was a huge pain to try and support all of those on the client side before converse. Converse just has one input and output schema. Some inference parameters have different upper and lower bounds or may not be supported on some models, but much easier to juggle

clayrosenthal:
On the [claude docs page](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html) they recommend converse:
> This topic shows how to use the Anthropic Claude messages API with the base inference operations ([InvokeModel](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModel.html) or [InvokeModelWithResponseStream](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModelWithResponseStream.html)). However, we recommend that you use the Converse API to implement messages in your application. The Converse API provides a unified set of parameters that work across all models that support messages. For more information, see [Carry out a conversation with the Converse API operations](https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html).

jeremychone:
@clayrosenthal Cool. Thanks for the feedback. 

By the way, we are not using any "persistent" features, right? Is everything stateless?

clayrosenthal:
yeah should be stateless, the client using this library would have to keep track of its own state of the chat and what not

PR #216, opened 2026-05-02
Add chat extra_body passthrough

zhangzhenhu:
This adds a generic chat request body extension point for provider-specific OpenAI-compatible fields.

  Changes:
  - Add `ChatOptions::with_extra_body(Value)`.
  - Add `ChatOptionsSet::extra_body()`.
  - Merge `extra_body` into OpenAI Chat Completions payloads.
  - Merge `extra_body` into OpenAI Responses payloads.
  - Treat streaming `usage:null` as absent usage instead of logging a deserialization error.
  - Add unit tests for Chat Completions payload merge, Responses payload merge, and null usage handling.

  Motivation:
  Some OpenAI-compatible providers expose useful non-standard request fields. For example, Aliyun/DashScope thinking models can require fields like `enable_thinking: false` to
  disable server-side thinking. Also, some compatible providers stream chunks with `usage:null`; this is a valid response shape and should not produce noisy deserialization
  errors.

### DIFF
--- a/src/adapter/adapters/openai/adapter_shared.rs
+++ b/src/adapter/adapters/openai/adapter_shared.rs
@@ -202,11 +202,21 @@ impl OpenAIAdapter {
 			}
 		}
 
+		// -- Provider-specific payload extension
+		// Merged last so callers can intentionally override previously set fields.
+		if let Some(extra_body) = options_set.extra_body() {
+			payload.x_merge(extra_body.clone())?;
+		}
+
 		Ok(WebRequestData { url, headers, payload })
 	}
 
 	/// Note: Needs to be called from super::streamer as well
 	pub(super) fn into_usage(adapter: AdapterKind, usage_value: Value) -> Usage {
+		if usage_value.is_null() {
+			return Usage::default();
+		}
+
 		// NOTE: here we make sure we do not fail since we do not want to break a response because usage parsing fail
 		let usage = serde_json::from_value(usage_value).map_err(|err| {
 			error!("Fail to deserialize usage. Cause: {err}");
@@ -489,10 +499,44 @@ struct OpenAIRequestParts {
 mod tests {
 	use super::*;
 	use crate::adapter::AdapterKind;
-	use crate::chat::{ChatMessage, ContentPart, MessageContent, ToolCall};
+	use crate::chat::{ChatMessage, ChatOptions, ContentPart, MessageContent, ToolCall};
 
 	fn test_model() -> ModelIden {
 		ModelIden::new(AdapterKind::OpenAI, "test-model")
+	}
+
+	#[test]
+	fn test_extra_body_merged_into_chat_completion_payload() {
+		let chat_options = ChatOptions::default()
+			.with_temperature(0.2)
+			.with_extra_body(json!({"temperature": 0.7, "enable_thinking": false}));
+		let options_set = ChatOptionsSet::default().with_chat_options(Some(&chat_options));
+		let target = ServiceTarget {
+			model: test_model(),
+			auth: AuthData::from_single("test-key"),
+			endpoint: Endpoint::from_static("https://api.openai.com/v1/"),
+		};
+
+		let web_req = OpenAIAdapter::util_to_web_request_data(
+			target,
+			ServiceType::Chat,
+			ChatRequest::from_user("hello"),
+			options_set,
+			None,
+		)
+		.expect("to_web_request_data should succeed");
+
+		assert_eq!(web_req.payload["enable_thinking"], false);
+		assert_eq!(web_req.payload["temperature"], 0.7);
+	}
+
+	#[test]
+	fn test_null_usage_is_treated_as_absent_usage() {
+		let usage = OpenAIAdapter::into_usage(AdapterKind::OpenAI, Value::Null);
+
+		assert!(usage.prompt_tokens.is_none());
+		assert!(usage.completion_tokens.is_none());
+		assert!(usage.total_tokens.is_none());
 	}
 
 	/// When an assistant message carries reasoning_content, it must appear
--- a/src/adapter/adapters/openai_resp/adapter_impl.rs
+++ b/src/adapter/adapters/openai_resp/adapter_impl.rs
@@ -249,6 +249,12 @@ impl Adapter for OpenAIRespAdapter {
 			}
 		}
 
+		// -- Provider-specific payload extension
+		// Merged last so callers can intentionally override previously set fields.
+		if let Some(extra_body) = chat_options.extra_body() {
+			payload.x_merge(extra_body.clone())?;
+		}
+
 		Ok(WebRequestData { url, headers, payload })
 	}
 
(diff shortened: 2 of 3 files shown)


jeremychone:
@zhangzhenhu Interesting. There are some PRs before that, but I will look into and think about this one. 

One question: Why only with OpenAI? In other words, why should `genai` assume that `with_extra_body(Value)` is only for OpenAI?

zhangzhenhu:
> @zhangzhenhu Interesting. There are some PRs before that, but I will look into and think about this one.
> 
> One question: Why only with OpenAI? In other words, why should `genai` assume that `with_extra_body(Value)` is only for OpenAI?



> @zhangzhenhu Interesting. There are some PRs before that, but I will look into and think about this one.
> 
> One question: Why only with OpenAI? In other words, why should `genai` assume that `with_extra_body(Value)` is only for OpenAI?

@jeremychone   Thanks, that is a fair question. I looked into this more carefully, and I think you are right to challenge the OpenAI-only implementation.

  What I found:

  - OpenAI's official Python SDK supports `extra_body` for Chat Completions and Responses as a way to add additional JSON properties to the request:
    - https://github.com/openai/openai-python/blob/main/src/openai/resources/chat/completions/completions.py
    - https://github.com/openai/openai-python/blob/main/src/openai/resources/responses/responses.py

  - Anthropic's official Python SDK also supports `extra_body` on `messages.create(...)`, with the same general purpose:
    - https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/resources/messages/messages.py

  - Google's Gen AI SDK has `HttpOptions.extra_body`, documented as extra request body parameters whose structure must match the backend API:
    - https://github.com/googleapis/python-genai
    - https://googleapis.github.io/js-genai/release_docs/interfaces/types.HttpOptions.html

  - Ollama native API is a little different: I did not find an `extra_body` escape hatch with that exact name, but its native chat API has provider-specific request controls such
  as `options` and `think`:
    - https://docs.ollama.com/api/chat
    - https://docs.ollama.com/capabilities/thinking

  So I agree that this should not be framed as an OpenAI-only concept.

  The reason I initially implemented it only for the OpenAI-compatible adapters was mostly historical/contextual, and honestly also some tunnel vision from the use case I was
  solving.

  In China, many model providers expose OpenAI-compatible endpoints as their primary integration path. Their non-standard features are often documented through the OpenAI SDK's
  `extra_body` escape hatch. Two concrete examples:

  - Aliyun/DashScope documents OpenAI-compatible Chat Completions thinking control via `extra_body={"enable_thinking": True}` because `enable_thinking` is not an OpenAI standard
  parameter:
    https://help.aliyun.com/zh/model-studio/deep-thinking

  - DeepSeek documents Thinking Mode for the OpenAI SDK via `extra_body={"thinking": {"type": "enabled"}}`:
    https://api-docs.deepseek.com/guides/thinking_mode

  That OpenAI-compatible ecosystem influenced my first implementation: the immediate validated use case was OpenAI-compatible providers, so I wired `extra_body` only into OpenAI
  Chat Completions and OpenAI Responses.

  In hindsight, that was too narrow for a generic `ChatOptions::with_extra_body(...)` API. If the option is generic, adapter support should either be clearly documented as
  explicit/partial, or the PR should implement it consistently for all adapters where the provider's official SDK/API supports an equivalent body-extension mechanism.

  I am happy to follow your preference here:

  1. If you prefer a smaller incremental PR, I can update this one to document that `extra_body` is a generic escape hatch but currently only supported by OpenAI-compatible
  adapters.

  2. If you prefer a consistent first version, I can close/replace this PR and submit a new one that adds support for all relevant adapters where the official provider SDK/API
  has an equivalent mechanism, including OpenAI, Anthropic, Gemini, and a carefully mapped approach for Ollama native.

  My preference after reviewing the provider SDKs is option 2, but I will defer to how you want genai's API surface to evolve.

jeremychone:
@zhangzhenhu btw, about the OpenAI vs others, I think it's fine to start with OpenAI for now as long as we know the intent it to add for others. This is probably the main usecase as you say. 

So, no need to add more in this PR for this.

jeremychone:
@zhangzhenhu Sorry, I deleted the comment about the `BEDROCK_API_KEY`. Not sure why my brain got confused. It's another PR.

jeremychone:
@zhangzhenhu So, what about I merge this one. Do an beta release so that you can use it. 

Then, you can do another PR for the other guys. 

What do you think?

zhangzhenhu:
> @zhangzhenhu So, what about I merge this one. Do an beta release so that you can use it.
> 
> Then, you can do another PR for the other guys.
> 
> What do you think?

@jeremychone   That sounds great to me.

  Merging this PR first and publishing a beta release would be very helpful for my side, because I can switch from the vendored patch back to an upstream genai version.

  Then I can open a follow-up PR for the other adapters, probably starting with Anthropic and Gemini since their official SDKs also expose an equivalent `extra_body` escape
  hatch. For Ollama native, I will look more carefully at the right mapping because its API uses native fields like `options` / `think` rather than the same `extra_body`
  convention.

  So yes, I am happy with this plan:
  1. merge this OpenAI-compatible implementation first;
  2. use the beta release downstream;
  3. prepare a separate PR for the other providers.

jeremychone:
@zhangzhenhu Awesome. I might merge some other PRs first, so, hopefully this will still pass. Plan to do those this weekend.

ropoctl:
@zhangzhenhu what do you think about making these extended supersets of openai completions api as one or more new adapters for ali and deepseek? seems like using typed parameters for thinking effort etc from the rust-genai would be more ergonomic than jamming json!{} literals into extra_body?

zhangzhenhu:
> @zhangzhenhu what do you think about making these extended supersets of openai completions api as one or more new adapters for ali and deepseek? seems like using typed parameters for thinking effort etc from the rust-genai would be more ergonomic than jamming json!{} literals into extra_body?

@ropoctl    I agree that typed parameters are the better ergonomic API for common/stable features.

  My intent with this PR is not to make `extra_body` the recommended way to control reasoning/thinking long term. For reasoning/thinking specifically, I think genai should prefer
  typed options where possible, and provider adapters can map those typed options to the provider-specific request shape.

  But `extra_body` still solves a different problem.

  Some OpenAI-compatible providers expose transitional or provider-specific fields before there is a stable cross-provider abstraction. Reasoning/thinking is one example, but not
  the only one. Third-party OpenAI-compatible providers often add non-standard request fields faster than SDKs can model them, and genai cannot realistically add typed support
  for every new provider field immediately.

  So I see the layers as:

  1. Typed genai options for common/stable behavior.
  2. Provider-specific adapter mappings for well-known providers like Aliyun or DeepSeek.
  3. `extra_body` as a low-level escape hatch when the provider exposes a field that genai has not modeled yet.

  In that sense, I do not see `extra_body` as competing with typed Aliyun/DeepSeek adapters. It is the fallback path that keeps users unblocked while the typed API catches up.

  So yes, I think adding dedicated typed mappings for Aliyun/DeepSeek can be a good follow-up. But I would still keep this PR focused on `extra_body`, because the escape hatch is
  generally useful for OpenAI-compatible providers beyond just reasoning/thinking.

jeremychone:
@zhangzhenhu @ropoctl I think we should keep this as Value, and we can have a `into<Value>` or serializable.

jeremychone:
extra commit message
```
Co-authored-by: zhangzhenhu <acmtiger@outlook.com>
```

jeremychone:
@zhangzhenhu Thanks, merged to main. Beta 19 release today or tomorrow.

PR #218, opened 2026-05-05
Simplify Gemini Streamer 

vagmi:
Gemini Streamer implements SSE. It is just poorly documented. 

https://github.com/google-gemini/cookbook/blob/main/quickstarts/rest/Streaming_REST.ipynb

Google's official clients actually use this query param but it is not documented on their API reference page.

Apologies for my earlier PR. Wanted to fix it in that but got caught up with other things.


ropoctl:
could you record a yakbak cassette that contains tool calls in the response?

jeremychone:
@vagmi This is really good work as usual. Thanks!

jeremychone:
@vagmi Much cleaner implementation of the Gemini stream than what I did. Big thanks!

vagmi:
Thanks for your kind words.

PR #220, opened 2026-05-06
gemini - add yakbak tool_stream cassette + replay test

ropoctl:
## Summary
- Records a streamed `functionCall` response from `gemini-3.1-flash-lite-preview` into `tests/data/yakbak/gemini/tool_stream/`.
- Adds replay test `test_yakbak_gemini_tool_stream` that asserts the SSE streamer captures the tool call (`get_weather` with `city=Paris`).
- Adds the matching `record_gemini_tool_stream` recorder helper.

Addresses the request on #218 to cover tool calls in the simplified Gemini SSE streamer. Kept as a separate cassette from `thinking_stream` since text and tool calls don't typically arrive in the same response.

## Test plan
- [x] `cargo test --test tests_yakbak_gemini` — both `thinking_stream` and `tool_stream` pass
- [x] `cargo build --tests` — compiles clean


jeremychone:
@ropoctl cool, will merge that tomorrow. I released v0.6.0-beta.19 will all of the previous PRs, some very good things there.

jeremychone:
pr commit addl comment: 

```md
* gemini - add yakbak tool_stream cassette + replay test

Records a streamed functionCall response from gemini-3.1-flash-lite-preview
and asserts the SSE streamer captures the tool call. Addresses the request
on PR #218 to cover tool calls in the simplified Gemini SSE streamer.

* gemini - tool_stream cassette - include thought signature

Re-record with reasoning_effort=High and capture_reasoning_content=true so
the cassette carries a `thoughtSignature` part next to the `functionCall`
(thoughtsTokenCount=114). Replay test now also asserts the captured thought
signature is forwarded and attached to the first tool call for handoff.

For Gemini 3.x the thought channel is the opaque `thoughtSignature`,
replacing the legacy `thought:true` + text shape used by 2.5.

* gemini - tool_stream cassette - thoughts + tool call together

Switch model from gemini-3.1-flash-lite-preview to gemini-3.1-pro-preview
and use a reasoning-heavy prompt so the cassette carries all four streams
in one response: visible text, `thought:true` summary text, a `functionCall`,
and a `thoughtSignature` blob.

The flash-lite-preview model only surfaces a `thoughtSignature` when a
function call is emitted (no `thought:true` parts). pro-preview emits
both, exercising every path in the SSE streamer in a single cassette.

Replay test now asserts captured content, captured reasoning text,
captured tool call (Cairo / C), and the thought signature carried on
the first tool call for handoff.
```

PR #227, opened 2026-05-13
openai_resp: round-trip encrypted reasoning blobs as input items

ropoctl:
## Problem

Reasoning models on the OpenAI Responses API (gpt-5.x, o-series, etc.) emit `type:"reasoning"` output items with an `encrypted_content` blob during streaming. To maintain reasoning state across turns — and to keep the provider-side prefix cache warm — those blobs need to ride back into the next request as top-level `type:"reasoning"` input items.

Today, `rust-genai` captures the blobs in some cases but drops them when serialising the next request, so the round-trip never closes. Also, some Responses-API-shaped backends (proxies that forward to internal endpoints, custom Azure deployments) deliver reasoning items *only* via `response.output_item.added` / `response.output_item.done` events and return an empty `response.output: []` at completion — so the existing capture path comes up empty there too.

### 1. Capture path (`streamer.rs`)

- Add a `RespStreamEvent::OutputItemDone` variant + handler that pulls `encrypted_content` off `type:"reasoning"` items into `captured_data.thought_signatures` as the items finalise.
- Keep the prior `response.completed.response.output` scan as a fallback (gated on `thought_signatures.is_none()`), so behaviour is unchanged on backends that populate `output`.
- Mark `ContentPartAdded`'s `_output_index` / `_content_index` as `#[serde(default)]`. They're sometimes absent on the wire and currently trip a parse warning + drop the event.

### 2. Replay path (`adapter_impl.rs`)

The assistant-message arm of the OpenAI Responses input builder silently dropped `ContentPart::ThoughtSignature` (and any `thought_signatures` stashed on a `ToolCall`). Now emits each blob as a top-level `{type: "reasoning", encrypted_content: <blob>, summary: []}` input item *before* the message / function_call items it belongs to, matching the order the API uses in its own streaming output.

## Tests

- New cassette `tests/data/yakbak/openai_resp/reasoning_stream_completed_empty/` reproduces the empty-`output` failure mode (synthesised from the existing `reasoning_stream` cassette by clearing `response.completed.response.output`).
- New `test_yakbak_openai_resp_reasoning_stream_completed_empty` asserts that `captured_thought_signatures()` still surfaces the blob from `output_item.done` in that case.
- Existing `test_yakbak_openai_resp_reasoning_stream` and all other yakbak tests in the file still pass — the `response.output` scan is preserved.

## Example

`examples/c11-openai-resp-reasoning-roundtrip.rs` walks through a two-turn conversation against the real OpenAI Responses API:
- Turn 1: ask a reasoning-triggering question, capture the encrypted blob via `StreamEnd::captured_thought_signatures()`.
- Turn 2: rebuild the assistant message with `ContentPart::ThoughtSignature(blob)` parts so the adapter emits the corresponding `type:"reasoning"` input item.
- Print `prompt_tokens_details.cached_tokens` each turn so the user can observe cache behaviour.

Run with `OPENAI_API_KEY=sk-... cargo run --example c11-openai-resp-reasoning-roundtrip`.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo check --examples`
- [x] `cargo test --test tests_yakbak_openai_resp` — 6/6 pass (5 pre-existing + 1 new)
- [ ] Live run of the new example against `api.openai.com` — recommend the maintainer try this with their own key before merging

### DIFF
--- a/examples/c11-openai-resp-reasoning-roundtrip.rs
+++ b/examples/c11-openai-resp-reasoning-roundtrip.rs
@@ -1,0 +1,169 @@
+//! Demonstrate the OpenAI Responses prefix-cache + reasoning round-trip.
+//!
+//! Reasoning models on the Responses API (gpt-5.x, o-series) emit
+//! `type:"reasoning"` items with an `encrypted_content` blob during
+//! streaming. To maintain reasoning state — and to keep the
+//! provider-side prefix cache warm — the next request in a
+//! conversation must replay those blobs as inputs alongside the
+//! assistant message they belong to.
+//!
+//! This example performs two turns:
+//!   1. Ask a reasoning-triggering question with `reasoning_effort=Medium`
+//!      and `capture_reasoning_content=true` (which causes the adapter
+//!      to set `include: ["reasoning.encrypted_content"]` on the request).
+//!   2. Re-send the conversation, attaching the captured
+//!      `ContentPart::ThoughtSignature` parts to the assistant turn.
+//!      The adapter emits each blob as a top-level `type:"reasoning"`
+//!      input item before the assistant message.
+//!
+//! Inspect `Usage.prompt_tokens_details.cached_tokens` on turn 2: when
+//! the backend honors prompt caching and the prefix matches, it will be
+//! non-zero. Cache-hit ratios vary by backend and conversation shape.
+//!
+//! Requires: OPENAI_API_KEY environment variable.
+//!
+//! Run: `OPENAI_API_KEY=sk-... cargo run --example c11-openai-resp-reasoning-roundtrip`
+
+use std::env;
+
+use futures::StreamExt;
+use genai::Client;
+use genai::ModelIden;
+use genai::adapter::AdapterKind;
+use genai::chat::{
+	ChatMessage, ChatOptions, ChatRequest, ChatStreamEvent, ContentPart, MessageContent, ReasoningEffort,
+};
+use genai::resolver::{AuthData, AuthResolver};
+
+/// A small reasoning-class model on the Responses API. Bump to a larger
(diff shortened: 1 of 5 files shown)


jeremychone:
@ropoctl I think something did this with Signature... on Gemini. Is that the same thing?

jeremychone:
@ropoctl Ho, that's cool, you use `ThoughtSignature`, so same mechanism than Gemini then, right?

ropoctl:
Yes — same abstraction. Reusing the existing `ContentPart::ThoughtSignature(String)` and `ToolCall.thought_signatures`, which the Gemini adapter already participates in on both ends.

The openai_resp adapter was the only piece missing: in the assistant arm of `adapter_impl.rs` it had `=> ()` no-ops for `ThoughtSignature` and `ReasoningContent` (old lines 455-456 / 501-502), silently dropping them on the way out. And the streamer captured encrypted reasoning by scanning `response.completed.response.output`, which some Responses-API-shaped backends return empty.

So the PR is mechanical:
1. **Capture (`streamer.rs`)**: new `OutputItemDone` handler reads `encrypted_content` from `type:"reasoning"` items into the *same* `captured_data.thought_signatures` field Gemini already populates. Keeps the old `response.output` scan as a fallback for backends that do echo `output` at completion.
2. **Serialize (`adapter_impl.rs`)**: emits `ContentPart::ThoughtSignature(blob)` and `ToolCall.thought_signatures` as top-level `{type:"reasoning", encrypted_content: blob, summary: []}` input items before the message/function_call items they belong to, mirroring the order the API itself emits them.

No new abstraction layer, same `ThoughtSignature` slot — openai_resp just now honours it on both ends like Gemini already does.

jeremychone:
@ropoctl This is awesome. Will merge this today.

jeremychone:
thanks for the PR, released in beta.20

PR #230, opened 2026-05-15
OpenRouter Support

xiao-e-yun:
This PR adds support for the OpenRouter adapter.

OpenRouter provides an OpenAI-compatible API (https://openrouter.ai/) with a wide range of models.

Features:
- Namespace-based routing: open_router::model-name
- OpenAI-compatible protocol (delegates to OpenAIAdapter)
- API key via environment variable OPEN_ROUTER_API_KEY
- Default endpoint: https://openrouter.ai/api/v1/

I'm a little confused about whether tests is mandatory.

### DIFF
--- a/docs/for-llm/api-reference-for-llm.md
+++ b/docs/for-llm/api-reference-for-llm.md
@@ -30,7 +30,7 @@ genai (crate root / lib.rs)
 - **ModelSpec**: Specifies a model at three resolution levels: `Name`, `Iden`, or `Target`.
 - **ServiceTarget**: Fully resolved call target: `ModelIden` + `Endpoint` + `AuthData`.
 - **Resolvers**: User hooks to customize model mapping, authentication, and service endpoints.
-- **AdapterKind**: Supported providers: `OpenAI`, `OpenAIResp`, `Gemini`, `Anthropic`, `Fireworks`, `Together`, `Groq`, `Mimo`, `Nebius`, `Xai`, `DeepSeek`, `Zai`, `BigModel`, `Aliyun`, `Cohere`, `Ollama`, `OllamaCloud`, `OpenCodeGo`, `Vertex`, `GithubCopilot`.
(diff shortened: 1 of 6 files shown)


xiao-e-yun:
Is it possible to change the OpenAI endpoint from hard-coded to `env.OPENAI_API_BASE ?? "https://api.openai.com/v1"`?

1. Modifying the endpoint in code, requires handling non-OpenAI adapters.

2. Implementing `serviceTargetResolver`, but this will tightly couple the provider to the code and makes `Namespacing` seem useless.

While this library supports multiple providers, providers not listed are a bit too cumbersome...?

I've already read #5 (which might be a bit outdated by 2024), but given that the vast majority of providers support OpenAI-compatible nodes, should we offer dedicated support for this use case?

jeremychone:
@xiao-e-yun First, thank you for this good PR. OpenRouter has been requested many times, but I never had the time to do it since I do not really use it.

> About the requirements/idea for env.OPENAI_API_BASE, there might be something there. We could do something like `[ADAPTER_NAME]_API_BASE_URL` env variables. This won't really add more adapters, but at least it allows changing the URL for a specific adapter. It could even be useful for Ollama (which now uses the native Ollama protocol).

> Anyway, we should open a discussion about this and think about it a little more. The trick is to see how it fits with everything else, ServiceTargetResolver and all.

xiao-e-yun:
The minimal modification probably only requires edit（OPENAI_API_BASE is OpenAI SDK default behaviour)

```diff 
 fn default_endpoint() -> Endpoint {
    const BASE_URL: &str = "https://api.openai.com/v1/";
-   Endpoint::from_static(BASE_URL) 
+   match std::env::var("OPENAI_API_BASE").ok() {
+      Some(url) => Endpoint::from_ownder(url),
+      None => Endpoint::from_static(BASE_URL)
+   }
 } 
```

https://github.com/jeremychone/rust-genai/blob/2ae40cd2c9fd84caad8686c8b3eccaf11c16a750/src/adapter/adapters/openai/adapter_impl.rs#L31-L34

If that's enough, I can create PR for this

jeremychone:
@xiao-e-yun Okay, but that should be a different discussion. I am not sure if we want to set it specifically for OpenAI like that.

This is not part of this PR, right?

xiao-e-yun:
Sorry for being off-topic, I made an issue for soliciting opinions, thank you very much for taking the time to reply.

jeremychone:
@xiao-e-yun thanks for the PR, release in beta.20

PR #232, opened 2026-05-21
Use the call_id returned by Gemini for function calling rather than synthesizing a new one.

vagmi:
I ran into a weird corner case where there are multiple tool calls for the same function but the responses where a bit unreliable. I wasn't able to reproduce the issue again but when I was going through the traces I noticed that the call ids generated by the Gemini inference service were ignored and we were synthesizing new ones. The documentation suggests that the id is optional but I always receive it. So I am using that and if that does not exist, I am using the synthesized one. I have recorded a yakbak cassette for this call as well.


jeremychone:
@vagmi THanks merged

PR #236, opened 2026-05-27
openai_resp: tolerate response.completed events without an output field

ropoctl:
## Summary

Some Responses-API-compatible backends omit the `output` field from the terminal `response.completed` SSE event entirely — not `output: []`, absent. Without `#[serde(default)]` on `RespResponse::output`, serde fails the whole event with `missing field \`output\``; the streamer logs a warn and skips it, then closes via the EOF fallback. The EOF fallback never finalises the per-`output_index` `in_progress_tool_calls` map into `captured_data.tool_calls`, so `captured_tool_calls()` returns `None` even though every `response.function_call_arguments.delta` arrived intact and the tool args are fully reassembled.

Fix is a one-liner on `RespResponse::output` plus a yakbak cassette + replay test recorded against a real backend where this reproduces. The cassette's `response.completed` payload has 33 keys — `output` is not one of them.

## Repro

Without the fix, the new test fails with:

```
Error: Tool call accumulated from delta events should survive a response.completed with no `output` field
```

With the fix, all 7 `tests_yakbak_openai_resp` tests pass.

## Test plan

- [x] `cargo test --test tests_yakbak_openai_resp` — 7 pass (was 6 pass + 1 new fail before the one-line fix)
- [x] `cargo check` clean
- [x] `cargo fmt --check` clean


jeremychone:
@ropoctl Merged and released as `0.6.2`

PR #239, opened 2026-05-31
build(reqwest): selectable TLS backends with a working rustls default

coaoac:
## Why

genai pins reqwest's `rustls` feature, which forces the **aws-lc-rs** crypto provider and pulls in `aws-lc-sys` — a C/assembly crate that needs aws-lc's kernel headers / libatomic and breaks (or badly complicates) lean **static-musl** and cross-compiled builds. Today there's no supported way to opt out without forking genai: a downstream shipping a static-musl service that standardizes on the `ring` provider had to fork just to keep `aws-lc-sys` out of its tree.

## What

Disable reqwest's default features and pin the non-TLS set (`json, stream, gzip, charset, http2, system-proxy`). Expose TLS via two genai features:

- **`rustls-tls`** (default) → `reqwest/rustls` = aws-lc-rs + OS trust store via `rustls-platform-verifier`. **Identical crypto to reqwest's own `default-tls`**, so existing users are unaffected and HTTPS works out of the box.
- **`native-tls`** → system OpenSSL / SChannel / Secure Transport.

Setting `default-features = false` drops the provider entirely, so a downstream can supply its own — e.g. `ring` via `CryptoProvider::install_default()` plus its own `reqwest` with `rustls-no-provider` — and keep `aws-lc-sys` off the tree. Custom root CAs / mTLS go through the existing `Client::builder().with_reqwest(...)`.

Adds a README **"TLS Backends"** decision table.

## Notes

- `rustls-tls` and `native-tls` are mutually exclusive at the reqwest level; select `native-tls` with `default-features = false`.
- Feature names use the ecosystem `-tls` convention — happy to rename to taste.
- Verified: the default (`rustls-tls`) and `--no-default-features` builds compile, and the replay test suite is green. `native-tls` wiring is correct (a local build just needs system OpenSSL).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- `!` build - reqwest TLS is now selectable via cargo features: `rustls-tls` (default; rustls + aws-lc-rs + OS trust store) and `native-tls`. reqwest switches to `default-features = false` with all previously-implicit features pinned explicitly; the default (`rustls-tls`) uses the same crypto backend as before, so HTTPS keeps working out of the box. BYO crypto provider / custom CA / mTLS via `ClientBuilder::with_reqwest`.
+- `!` build - reqwest TLS is now selectable via cargo features: `rustls-tls` (default; rustls + aws-lc-rs + OS trust store) and `native-tls`. reqwest switches to `default-features = false` with all previously-implicit features pinned explicitly; the default (`rustls-tls`) uses the same crypto backend as before, so HTTPS keeps working out of the box. BYO crypto provider / custom CA / mTLS via `ClientBuilder::with_reqwest`. Enabling both backends at once fails fast with a `compile_error!` instead of silently picking one.
 
 ## 2026-05-29 - [v0.6.1](https://github.com/jeremychone/rust-genai/compare/v0.6.2...v0.6.3)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 `.` minor | `-` Fix | `+` Addition | `^` improvement | `!` Change | `>` Refactor
 
+## Unreleased
+
+- `!` build - reqwest TLS is now selectable via cargo features: `rustls-tls` (default; rustls + aws-lc-rs + OS trust store) and `native-tls`. reqwest switches to `default-features = false` with all previously-implicit features pinned explicitly; the default (`rustls-tls`) uses the same crypto backend as before, so HTTPS keeps working out of the box. BYO crypto provider / custom CA / mTLS via `ClientBuilder::with_reqwest`.
+
 ## 2026-05-29 - [v0.6.1](https://github.com/jeremychone/rust-genai/compare/v0.6.2...v0.6.3)
 
 - `^` use adaptive thinking for opus 4.7 (and above)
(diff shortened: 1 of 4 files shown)


jeremychone:
@coaoac So, to make sure. If I do not select any features, it will work like before, right ?

jeremychone:
Looks very good addition btw. Thanks for this great PR.

coaoac:
Hello! 
Default behavior is unchanged — with default features you get the same rustls + aws-lc-rs + OS-trust-store backend as before (reqwest switches to default-features = false but every previous default is re-added explicitly).
Two edge cases:
1. `default-features = false` on genai now drops the TLS backend (it was a no-op before). This is intentional — it's exactly the bring-your-own-client / ring / musl path the README documents (wire up reqwest's TLS yourself or inject a client via with_reqwest).
2. Enabling `native-tls` without `default-features = false` used to leave both backends compiled, with reqwest silently picking one. I added a new guard that turns that into a clear build error.

jeremychone:
Git addl comment:
```
* build(reqwest): selectable TLS backends with a working rustls default

genai pinned reqwest's `rustls` feature, forcing the aws-lc-rs crypto provider
and pulling in aws-lc-sys (a C/asm crate needing aws-lc kernel headers/libatomic),
which breaks lean static-musl and cross-compiled builds — with no supported way
to opt out short of forking genai.

Disable reqwest default features and pin the non-TLS set
(json, stream, gzip, charset, http2, system-proxy). Expose TLS via two genai
features: `rustls-tls` (default; reqwest/rustls = aws-lc-rs + OS trust store via
rustls-platform-verifier) and `native-tls`. The default crypto is identical to
reqwest's own default-tls, so existing users are unaffected and HTTPS works out
of the box. `default-features = false` drops the provider, so a downstream can
supply its own (e.g. ring via CryptoProvider::install_default() + its own reqwest
with rustls-no-provider) and keep aws-lc-sys off the tree; custom CAs / mTLS go
through the existing Client::builder().with_reqwest(...).

Adds a README "TLS Backends" decision table.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

* build(reqwest): reject enabling both TLS backends at compile time

`rustls-tls` (default) and `native-tls` are mutually exclusive. Enabling
`native-tls` without `default-features = false` left `rustls-tls` on from the
default set, so both reqwest backends compiled and reqwest silently picked one
— a quiet mis-selection.

Add a `compile_error!` in src/lib.rs gated on `all(rustls-tls, native-tls)` that
turns that into a clear, actionable build error. The "neither feature" case is
intentionally left compilable: it is the supported bring-your-own-client path
via `with_reqwest`. Update the Cargo.toml/README/CHANGELOG notes that previously
described the silent behavior.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

---------

Co-authored-by: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
```

jeremychone:
@coaoac feel free to let me know if you would like a `0.6.4` with this. I can make that release sooner if needed

coaoac:
> @coaoac feel free to let me know if you would like a `0.6.4` with this. I can make that release sooner if needed

Thanks but no hurry. I can keep my fork until then and might find other pull request opportunities in the coming days.

PR #240, opened 2026-05-31
feat(anthropic): prompt caching on tools + request-level cache control

coaoac:
## Why

Anthropic prompt caching makes cache reads ~10× cheaper than normal input tokens and cuts latency when an app re-sends a large, stable prefix — tool definitions and system prompts — on every request (agents, RAG, long-system-prompt chatbots). genai couldn't fully exploit it:

- **tool definitions had no way to be cached**, and
- **request-level `ChatOptions::with_cache_control` was silently ignored on Anthropic** (logged, then dropped) — so the most ergonomic "just cache my stable prefix" knob did nothing.

This PR closes both gaps.

## What

- **`Tool::with_cache_control(...)`** — serialized onto the Anthropic `tools` payload, so the tool-definition prefix can be cached.
- **Request-level `ChatOptions::with_cache_control(...)`** now auto-marks the static prefix on Anthropic — the last `system` block (caches tools+system in canonical order), or the last tool when there's no system — **only** when no explicit message/tool breakpoint exists. It never overwrites an explicit TTL; no static prefix → no-op. So `ChatOptions::default().with_cache_control(CacheControl::Ephemeral)` now caches the stable tools+system prefix instead of doing nothing.
- Documents that **`CacheControl::Ephemeral24h` clamps to Anthropic's max `1h`** TTL (it still maps to a real 24h on providers that support it, e.g. OpenAI `prompt_cache_retention`).
- `into_anthropic_request_parts` gained a `request_cache_control` param, so the `vertex` and `opencode_go` call sites (both Anthropic-protocol) were updated and consistently gain request-level caching.

## Tests

4 unit tests: system auto-mark, tool fallback, no-op-without-prefix, and defer-to-explicit. `cargo test --lib`, `cargo clippy --all-targets`, and the yakbak replay suite pass.

---

*Rebased onto current `main` (`0.7.0-beta.2-WIP`): after the `anthropic_shared` refactor, the request-building logic and these cache changes now live in `adapter_shared.rs` (`adapter_impl.rs` unchanged). `cargo test --lib` and `cargo clippy --all-targets` pass.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 `.` minor | `-` Fix | `+` Addition | `^` improvement | `!` Change | `>` Refactor
 
+## Unreleased
+
+- `+` anthropic - prompt caching on tools via `Tool::with_cache_control`, and request-level `ChatOptions::with_cache_control` now auto-applies a cache breakpoint to the static (tools+system) prefix (was previously ignored). `Ephemeral24h` is documented as clamped to Anthropic's max `1h` TTL.
+
 ## 2026-06-06 [v0.6.5](https://github.com/jeremychone/rust-genai/compare/v0.6.4...v0.6.5)
 
 - `-` fix: capture OpenAI stream usage tail (PR #242)
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Here’s what’s new:
 - **Chat extra body**: `ChatOptions::with_extra_body(...)` provides a low-level request body extension point for provider-specific fields in OpenAI-compatible chat payloads.
 - **Tool choice**: `ChatOptions::with_tool_choice(...)` adds provider-neutral tool selection hints for automatic, disabled, required, or specific tool calls.
 - **Built-in tools and WebSearch**: Added typed built-in tool support, including `ToolName`, `ToolConfig`, `WebSearch`, and provider mappings for Anthropic, OpenAI, and Gemini.
-- **Prompt cache controls**: Chat-level `CacheControl` support adds provider-specific prompt caching options, including OpenAI `prompt_cache_key` and cache retention.
+- **Prompt cache controls**: Chat-level `CacheControl` support adds provider-specific prompt caching options, including OpenAI `prompt_cache_key` and cache retention. On Anthropic, `Tool::with_cache_control` marks individual tools, and request-level `ChatOptions::with_cache_control` auto-applies a breakpoint to the static (tools+system) prefix.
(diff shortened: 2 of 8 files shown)


jeremychone:
@coaoac This is good. 

Since it introduces an edge-breaking change (if the user destructures `Tool`), I think I want to put this in a 0.7.0. 

So, I am planning to wait a little for that one, to see if there are other API changes. 

Does that work for you, or would you like this sooner rather than later?

coaoac:
Sounds good. Thanks!

jeremychone:
@coaoac Sorry, I added a minimax provider, and since it uses Anthropic (recommended), I used this opportunity to make the anthropic `adapter_shared.rs` follow the openai pattern. 

So, now, this PR has some conflicts.

But this was a very good PR. 

I can run the `pro@coder` on it and push a temp branch for you to check and redo a PR if you want. 

This is definitely a good feature to have, so I do not want to lose it.

coaoac:
Ok no problem. I can do a new PR. Let me know when the branch is ready

jeremychone:
@coaoac Ok, the main branch is ready, now `0.7.0-beta.1-WIP`

You can redo this PR. 

BTW, side note: in `v0.7.0` I added some cool new custom adapters that get triggered with namespaces. `genai_1::some_model`, this will look in the env `GENAI_1_ENDPOINT` and `GENAI_1_API_KEY`.

jeremychone:
Addl commit info

```
Anthropic prompt caching makes cache reads ~10x cheaper than normal input tokens
and cuts latency when an app re-sends a large, stable prefix (tool definitions,
system prompts) on every request. genai couldn't fully exploit it: tools had no
cache_control, and request-level ChatOptions::with_cache_control was silently
ignored on Anthropic (logged, then dropped).

- Add Tool::with_cache_control(...), serialized onto the Anthropic tools payload.
- Request-level ChatOptions::with_cache_control now auto-marks the static prefix
  (last system block -> caches tools+system; else the last tool) only when no
  explicit message/tool breakpoint is set; never overwrites an explicit TTL;
  no static prefix -> no-op.
- Document that CacheControl::Ephemeral24h clamps to Anthropic's max 1h TTL.
- into_anthropic_request_parts gained a request_cache_control param; vertex and
  opencode_go call sites (both Anthropic-protocol) updated and consistently gain
  request-level caching.
- 4 unit tests.

Co-authored-by: Claude Opus 4.8 (1M context) <noreply@anthropic.com>
```

jeremychone:
@coaoac released in `0.7.0-beta.2`

PR #242, opened 2026-06-06
fix: capture OpenAI stream usage tail

### LINKED ISSUE #241: OpenAI-compatible streamer drops usage when finish_reason chunk has usage:null before usage-only tail
## Summary

For some OpenAI-compatible streaming APIs with `stream_options.include_usage=true`, the SSE stream sends:

1. a normal final choice chunk with `choices[0].finish_reason` and `usage: null`
2. a later usage-only tail chunk with `choices: []` and a real `usage` object
3. `[DONE]`

With the current OpenAI streamer, `ChatStreamEvent::End.captured_usage` is present but empty (`prompt_tokens=None`, `completion_tokens=None`, `total_tokens=None`), so downstream callers cannot read provider usage even though the SSE response contains it.

## Observed SSE tail shape

Sanitized example from an OpenAI-compatible endpoint:

```json
{"choices":[{"finish_reason":"length","delta":{}}],"usage":null}
{"choices":[],"usage":{"prompt_tokens":259,"completion_tokens":16,"total_tokens":275}}
[DONE]
```

## Reproduction result

Using rust-genai at rev `ea79a86` with `ChatOptions::default().with_capture_usage(true)` against that endpoint:

```text
adapter=deepseek captured_usage_present=true prompt_tokens=None completion_tokens=None total_tokens=None
adapter=openai captured_usage_present=true prompt_tokens=None completion_tokens=None total_tokens=None
```

The same raw request made directly over HTTP confirms the usage-only tail frame contains `prompt_tokens=259`.

## Suspected cause

In `src/adapter/adapters/openai/streamer.rs`:

- `take_finish_reason_usage` receives the finish_reason frame.
- When that frame has `usage: null`, `OpenAIAdapter::into_usage(Value::Null)` returns `Usage::default()`.
- For provider-specific branches like DeepSeek/Zai/Fireworks/Together, `unwrap_or_default()` wraps that empty usage in `Some(...)`.
- Later, the usage-only tail frame is only captured when `self.captured_data.usage.is_none()`.
- Because the previous `usage:null` already set `captured_data.usage = Some(default_usage)`, the real tail usage is skipped.

Additionally, the usage-only tail path excludes `AdapterKind::DeepSeek`, so DeepSeek-compatible endpoints that place real usage in the tail cannot be captured there.

Relevant logic in current `main` still appears to have:

```rust
AdapterKind::DeepSeek | AdapterKind::Zai | AdapterKind::Fireworks | AdapterKind::Together => Some(
    message_data
        .x_take("usage")
        .map(|v| OpenAIAdapter::into_usage(adapter_kind, v))
        .unwrap_or_default(),
),
```

and later:

```rust
if !matches!(adapter_kind, AdapterKind::Groq)
    && !matches!(adapter_kind, AdapterKind::DeepSeek)
    && self.captured_data.usage.is_none()
    && self.options.capture_usage
{
    ... capture usage-only tail ...
}
```

## Expected behavior

`usage:null` should not occupy `captured_data.usage` as a successful usage capture. A later usage-only tail frame with real usage should be allowed to overwrite an empty/default captured usage, or the streamer should not store empty usage from null in the first place.

For the sample SSE above, `End.captured_usage.prompt_tokens` should be `Some(259)`.

## Notes

I found #186, which fixed inline real usage on `finish_reason` chunks. This looks like a related but different shape: the finish_reason chunk has `usage:null`, and real usage arrives in the next usage-only chunk.

kawayiYokami:
Fixes #241.

The official Mimo API can send usage separately from the finish_reason chunk in an OpenAI-compatible stream:

1. a finish_reason chunk with `usage: null` or no usable token counts
2. a final usage-only chunk with `choices: []` and the real `usage` object

Before this change, the streamer could capture null or empty parsed usage as an empty `Usage` value. That empty value then blocked the later usage-only tail chunk from being recorded, so token usage was lost for this response shape.

This change:
- treats null or empty parsed usage as absent, so it does not block later real usage
- captures usage-only tail chunks when they contain real token counts
- applies this to the OpenAI-compatible streaming path, including Mimo official API responses
- keeps Groq's provider-specific `x_groq.usage` handling unchanged
- preserves already-captured non-empty usage

Mimo official API is the provider where this issue was observed. The fix is intentionally implemented in the shared OpenAI-compatible streamer because the response shape is not Mimo-specific.

Tests:
- `cargo fmt --check`
- `cargo test --lib`

### DIFF
--- a/src/adapter/adapters/openai/streamer.rs
+++ b/src/adapter/adapters/openai/streamer.rs
@@ -2,7 +2,7 @@ use crate::adapter::AdapterKind;
 use crate::adapter::adapters::support::{StreamerCapturedData, StreamerOptions};
 use crate::adapter::inter_stream::{InterStreamEnd, InterStreamEvent};
 use crate::adapter::openai::OpenAIAdapter;
-use crate::chat::{ChatOptionsSet, StopReason, ToolCall};
+use crate::chat::{ChatOptionsSet, StopReason, ToolCall, Usage};
 use crate::webc::{Event, EventSourceStream};
 use crate::{Error, ModelIden, Result};
 use serde_json::Value;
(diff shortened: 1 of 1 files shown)


jeremychone:
@kawayiYokami thanks looks very good. Let me do some sanity check, and will merge that one. 

Thanks a lot!

jeremychone:
Released in `v0.6.5`

PR #243, opened 2026-06-07
Derive PartialEq/Eq for usage

Alb-O:
Add PartialEq and Eq derives to Usage and its nested structs so we can compare usage values in downstream tests.


jeremychone:
@Alb-O Thanks, merged, will be in the upcoming `0.7.0-beta.2`

PR #247, opened 2026-06-16
feat(anthropic): support fine-grained tool streaming (eager_input_streaming)

coaoac:
## Summary

Adds support for Anthropic's **fine-grained tool streaming** (now GA), exposed as an opt-in per-tool flag.

- `Tool` gains `eager_input_streaming: Option<bool>` + builder `with_eager_input_streaming(bool)` (mirrors the existing `strict` / `cache_control` fields).
- The Anthropic adapter emits `"eager_input_streaming": true` on a custom tool when the flag is set.

## Why

By default Anthropic buffers tool-call input server-side and emits it in coarse `input_json_delta` grains. With `eager_input_streaming` enabled, the provider streams tool-call input token-granular, so consumers can render tool arguments progressively as they arrive (e.g. live UI that reflects a tool's parameters as the model writes them). This is the GA, per-tool successor to the old `anthropic-beta: fine-grained-tool-streaming-2025-05-14` header — no beta header required.

Docs: https://platform.claude.com/docs/en/agents-and-tools/tool-use/fine-grained-tool-streaming

## Notes

- Opt-in and serialized only when `Some(true)`, so existing behavior is unchanged.
- Applied to custom tools (the `input_schema` branch); built-in tools are untouched.

## Test Plan

- [x] `test_tool_eager_input_streaming_serializes` — `with_eager_input_streaming(true)` round-trips through `to_web_request_data` and the serialized tool carries `eager_input_streaming: true`.
- [x] `cargo test --lib` (112 passing), `cargo fmt --check` clean.

### DIFF
--- a/src/adapter/adapters/anthropic/adapter_shared.rs
+++ b/src/adapter/adapters/anthropic/adapter_shared.rs
@@ -640,6 +640,7 @@ impl AnthropicAdapter {
 			schema,
 			config,
 			cache_control,
+			eager_input_streaming,
 			..
 		} = tool;
 
@@ -686,6 +687,10 @@ impl AnthropicAdapter {
 			if let Some(description) = description {
 				// TODO: need to handle error
 				let _ = tool_value.x_insert("description", description);
+			}
+			// Anthropic fine-grained tool streaming (GA): opt-in per tool.
+			if eager_input_streaming == Some(true) {
+				let _ = tool_value.x_insert("eager_input_streaming", true);
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


jeremychone:
PR commit comment

```
Add Tool::eager_input_streaming (Option<bool>) + with_eager_input_streaming(bool); the
Anthropic adapter emits `eager_input_streaming: true` on a custom tool when set. This is
Anthropic's GA fine-grained tool streaming (per-tool opt-in, no beta header) — the provider
streams tool-call input without buffering the full JSON, so consumers can render tool
arguments progressively as input_json_delta arrives.

```

jeremychone:
released in v0.7.0-beta.4

PR #249, opened 2026-06-20
fix: reuse Client WebClient for model listing

### LINKED ISSUE #248: Threading WebClient to model listing
`with_reqwest(...)` is documented as the way to install custom root CAs etc. Chat, stream, and embed use the client’s configured transport via `self.web_client()`.

Model listing does not. `Client::all_model_names` resolves endpoint/auth, then calls:

```rust
AdapterDispatcher::all_model_names(adapter_kind, endpoint, auth).await?
```

The dispatcher and `Adapter::all_model_names` trait carry no `WebClient`, so the live HTTP listing paths create their own `WebClient::default()` instead (OpenAI-compatible shared listing, Anthropic, Gemini, and Ollama).

That makes `with_reqwest(...)` incomplete. For example, in a (very) bare container with no system trust store this can panic during startup provider discovery, because `WebClient::default()` builds a default rustls/platform-verifier reqwest client and `.expect(...)`s the build result.

**Expected behavior**: every HTTP request made by a `Client` should use that Client’s configured `WebClient`.

**Proposed fix**: Similar to #155/#156, thread `&WebClient` from `Client::all_model_names` into  `AdapterDispatcher::all_model_names`, `Adapter::all_model_names`, etc.

lambdabetaeta:
## Summary
- Thread `Client`'s configured `WebClient` through adapter model-listing calls.
- Remove ad-hoc `WebClient::default()` creation in OpenAI, Anthropic, Gemini, and Ollama model-listing paths.
- Preserve static model-list providers while accepting the shared web client parameter.

Fixes #248

### DIFF
--- a/src/adapter/adapter_types.rs
+++ b/src/adapter/adapter_types.rs
@@ -2,7 +2,7 @@ use crate::adapter::AdapterKind;
 use crate::chat::{ChatOptionsSet, ChatRequest, ChatResponse, ChatStreamResponse};
 use crate::embed::{EmbedOptionsSet, EmbedRequest, EmbedResponse};
 use crate::resolver::{AuthData, Endpoint};
-use crate::webc::WebResponse;
+use crate::webc::{WebClient, WebResponse};
 use crate::{Headers, ModelIden};
 use crate::{Result, ServiceTarget};
 use reqwest::RequestBuilder;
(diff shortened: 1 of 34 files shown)


jeremychone:
@lambdabetaeta thank you, all merged. 

> I am going to add kimi as provider, and then do a release.

PR #250, opened 2026-07-01
fix: support adaptive thinking for Claude Sonnet 5

lambdabetaeta:
## Summary

Adds Claude Sonnet 5 to the Anthropic adapter’s adaptive-thinking capability checks.

Sonnet 5 uses the newer adaptive thinking request shape:

- `thinking: { "type": "adaptive" }`
- `output_config.effort` for reasoning effort control
- `max` and `xhigh` effort support

This updates the model capability gates so `claude-sonnet-5` follows the same adaptive-thinking path as the currently supported Claude 4.6+ adaptive models, and adds a regression test for the generated request payload.

## Testing

- `cargo test adapter::adapters::anthropic::adapter_shared::tests::test_sonnet_5_uses_adaptive_thinking`


jeremychone:
@lambdabetaeta thanks, good catch, merged. 

I merged it, and then, created a `support...xhigh(model_name)` for symmetry with the other support. 3c8028f2a3df0159f998db352d8126fda652ace3

Later, I will probably move all of those into a bitflags of some sort.

PR #253, opened 2026-07-04
fix: ReasoningEffort::Zero (rename from ::None) disables reasoning on Anthropic (#251)

### LINKED ISSUE #251: ReasoningEffort::None Enables Anthropic Adaptive Thinking
# `ReasoningEffort::None` Enables Anthropic Adaptive Thinking

## Summary

The Anthropic adapter currently treats `ReasoningEffort::None` inconsistently for models that support adaptive thinking. A caller can explicitly request no reasoning, but the generated Anthropic payload may still enable adaptive thinking.

This affects adaptive-thinking models such as Claude Opus 4.7+ and Claude Sonnet 5. The exact expected payload differs by model because Anthropic uses different default behavior for these model families.

## Overview

Anthropic reasoning payloads are built in `insert_anthropic_reasoning`. When an adaptive-capable model is used, the function eventually inserts:

```json
{
  "thinking": {
    "type": "adaptive"
  }
}
```

That happens whenever `support_adaptive` is true.

Before that, `ReasoningEffort::None` is handled only by suppressing `output_config.effort`:

```rust
ReasoningEffort::None => "",
```

The empty string prevents this field from being emitted:

```json
{
  "output_config": {
    "effort": "..."
  }
}
```

However, it does not stop the later adaptive-thinking block from running. Thus an explicit `ReasoningEffort::None` can still produce:

```json
{
  "thinking": {
    "type": "adaptive"
  }
}
```

while `ReasoningEffort::None` is expected to mean positively no reasoning effort.

## Expected Behavior

`ReasoningEffort::None` should disable reasoning for the target model.

For Claude Opus 4.7+ omitting the `thinking` field is enough. Anthropic documents adaptive thinking for these models as opt-in: thinking is off unless the request explicitly includes:

```json
{
  "thinking": {
    "type": "adaptive"
  }
}
```

Therefore, for Opus 4.7+ with `ReasoningEffort::None`, the adapter should emit no `thinking` field and no `output_config.effort`.

Claude Sonnet 5 is different: adaptive thinking as on by default. To turn it off, the
request must explicitly include:

```json
{
  "thinking": {
    "type": "disabled"
  }
}
```

Therefore, for Sonnet 5 with `ReasoningEffort::None`, the adapter should emit `thinking: {"type": "disabled"}` and no `output_config.effort`.

## Suggested Fix

Handle `ReasoningEffort::None` before the generic adaptive-thinking serialization
path.

One possible implementation:

```rust
if matches!(effort, ReasoningEffort::None) {
    if model_name.contains("claude-sonnet-5") {
        payload.x_insert("thinking", json!({"type": "disabled"}))?;
    }
    return Ok(());
}
```

This should be placed near the top of `insert_anthropic_reasoning`, before the
`support_effort` and `support_adaptive` branches insert any reasoning fields.

The Sonnet 5 condition may be worth extracting into a named helper if more models
adopt the same "adaptive thinking is on by default" behavior later.

## Suggested Tests

Add regression tests for both expected behaviors:

- `claude-opus-4-7` with `ReasoningEffort::None`
  - does not include `thinking`
  - does not include `output_config.effort`

- `claude-sonnet-5` with `ReasoningEffort::None`
  - includes `thinking: {"type": "disabled"}`
  - does not include `output_config.effort`

coaoac:
Fixes #251 — implements the rename agreed there.

## Problem

`Some(ReasoningEffort::None)` suppressed `output_config.effort` but still fell through to the adaptive-thinking branch, so an explicit "no reasoning" request emitted `thinking: {"type": "adaptive"}` on adaptive-capable models. On Claude Sonnet 5 — where adaptive thinking is on by default when the `thinking` field is omitted — there was no way to turn thinking off at all.

## Change

**1. `ReasoningEffort::None` → `ReasoningEffort::Zero`** (as proposed in #251), so the variant no longer reads like `Option::None`. The keyword surface is unchanged: `as_keyword()`/`Display` still produce `"none"`, `from_keyword()` still parses `"none"`, so the `-none` model suffix keeps working everywhere. `variant_name()` returns `"zero"` (its contract is the variant name; the keyword-vs-variant split is intentional). `#[serde(alias = "None")]` keeps previously serialized values deserializable — note the serialize side now emits `"Zero"`.

**2. Anthropic: `Zero` now positively disables reasoning.** `insert_anthropic_reasoning` handles it before the effort/adaptive/legacy branches:

| Model family | Payload for `Zero` | Why |
|---|---|---|
| Claude Sonnet 5 | `thinking: {"type": "disabled"}` | adaptive thinking is on by default; omitting the field would leave it on |
| Opus 4.7 / 4.8, Opus 4.6 / Sonnet 4.6, legacy | *(no `thinking`, no `output_config.effort`)* | thinking is off unless requested |
| Fable / Mythos | *(no `thinking`)* | thinking is always-on and an explicit `"disabled"` is rejected — omitting is "as zero as possible" |

The "on by default" set lives in a named helper (`anthropic_thinking_on_by_default`) per the suggestion in #251, ready for future models that adopt the same default.

**3. Anthropic model-suffix table:** `-zero` and `-none` now both map to `Zero` and are stripped from the model name. Two behavior deltas worth calling out explicitly:
- previously `"zero" => None` matched but didn't strip (the literal `claude-sonnet-5-zero` was sent to the API);
- the `"None" => Some(Low)` entry (capital N, mapping a "disable"-looking suffix to *low reasoning*) is removed rather than renamed — it looked like a leftover; shout if it was intentional.

This is scoped to the Anthropic adapter's own table — the shared `from_keyword()` deliberately does **not** learn `"zero"`, since it feeds `from_model_name()` in the OpenAI-family adapters where real model ids end in `-zero` (e.g. `deepseek-r1-zero`) and would be silently rewritten.

Other adapters (bedrock, gemini, openai) are mechanical renames; their existing behavior for this variant is unchanged (openai already maps it to OpenAI's native `reasoning_effort: "none"`).

## Not addressed here (pre-existing, happy to file/follow up)

- **Gemini**: `Zero` still emits no `thinkingConfig`, so thinking-by-default Gemini models keep thinking. Gemini has a positive off (`thinkingBudget: 0` — what the Gemini `-zero` suffix already maps to via `Budget(0)`); mapping `Zero` to it would mirror this fix but is beyond #251's scope.
- **Bedrock**: `Zero` emits no fields (unchanged), and the Anthropic-on-Bedrock path still uses the legacy `thinking.enabled + budget_tokens` shape for non-Zero efforts, which newer models reject.

## Tests

The two regression tests suggested in #251, plus suffix and other-model pins:

- `test_opus_4_7_reasoning_zero_omits_thinking` — no `thinking`, no `output_config`
- `test_sonnet_5_reasoning_zero_disables_thinking` — `thinking: {"type":"disabled"}`, no `output_config`
- `test_reasoning_zero_model_name_suffix` — `claude-sonnet-5-none` / `-zero` strip the suffix and disable thinking
- `test_reasoning_zero_other_models_omit_thinking` — legacy (`claude-sonnet-4-5`), off-by-default adaptive (`claude-opus-4-6`), and `claude-fable-5` (pins the "never send disabled to Fable" invariant from the helper's NOTE)

`cargo test --lib`: 117 passed; clippy clean. (The two doctest failures in `tool_base.rs` / `zai/mod.rs` are pre-existing on main.)

### DIFF
--- a/src/adapter/adapters/anthropic/adapter_shared.rs
+++ b/src/adapter/adapters/anthropic/adapter_shared.rs
@@ -419,13 +419,12 @@ impl AnthropicAdapter {
 
 		// -- Reasoning Budget
 		let (model_name, computed_reasoning_effort) = match (raw_model_name, options_set.reasoning_effort()) {
-			// No explicity reasoning_effor, try to infer from model name suffix (supports -zero)
+			// No explicity reasoning_effor, try to infer from model name suffix (supports -zero and -none)
 			(model, None) => {
 				// let model_name: &str = &model.model_name;
 				if let Some((prefix, last)) = raw_model_name.rsplit_once('-') {
 					let reasoning = match last {
-						"zero" => None,
-						"None" => Some(ReasoningEffort::Low),
+						"zero" | "none" => Some(ReasoningEffort::Zero),
 						"minimal" => Some(ReasoningEffort::Low),
 						"low" => Some(ReasoningEffort::Low),
 						"medium" => Some(ReasoningEffort::Medium),
@@ -730,6 +729,16 @@ fn insert_anthropic_reasoning(
 	let support_adaptive = supports_anthropic_adaptive_thinking(model_name);
 	let support_xhigh = supports_anthropic_reasoning_xhigh(model_name);
 
+	// `Zero` means "no reasoning": emit no `output_config.effort` and no adaptive/legacy
+	// thinking payload. Models where thinking is on by default need an explicit opt-out;
+	// for all others, omitting the `thinking` field already means "off".
+	if matches!(effort, ReasoningEffort::Zero) {
+		if anthropic_thinking_on_by_default(model_name) {
+			payload.x_insert("thinking", json!({"type": "disabled"}))?;
+		}
+		return Ok(());
+	}
+
 	// Models that support effort use it as the primary reasoning control.
 	if support_effort {
 		let effort = match effort {
@@ -747,7 +756,8 @@ fn insert_anthropic_reasoning(
 				budget = Some(*val);
 				""
 			}
-			ReasoningEffort::None => "",
+			// Handled by the early return above; kept for exhaustiveness.
+			ReasoningEffort::Zero => "",
 		};
 
 		// Emit the effort into the shared output_config map when present.
@@ -774,7 +784,7 @@ fn insert_anthropic_reasoning(
 	// Older models still use the legacy `thinking.enabled + budget_tokens` shape.
 	if !support_effort {
 		let thinking_budget = match effort {
-			ReasoningEffort::None => None,
+			ReasoningEffort::Zero => None,
 			ReasoningEffort::Budget(budget) => Some(*budget),
 			ReasoningEffort::Low | ReasoningEffort::Minimal => Some(REASONING_LOW),
 			ReasoningEffort::Medium => Some(REASONING_MEDIUM),
@@ -815,6 +825,16 @@ fn supports_anthropic_reasoning_max(model_name: &str) -> bool {
 
 fn supports_anthropic_reasoning_xhigh(model_name: &str) -> bool {
 	is_opus_4_7_or_higher(model_name) || model_name.contains("claude-sonnet-5")
+}
+
+/// Models where Anthropic enables adaptive thinking by default when the request omits the
+/// `thinking` field (currently the Claude Sonnet 5 family). These need an explicit
+/// `{"type": "disabled"}` to turn reasoning off.
+///
+/// NOTE: Fable/Mythos thinking is always-on and cannot be disabled (an explicit "disabled"
+/// is rejected), so they are intentionally not listed — for them, `Zero` omits `thinking`.
+fn anthropic_thinking_on_by_default(model_name: &str) -> bool {
+	model_name.contains("claude-sonnet-5")
 }
 
 fn supports_anthropic_adaptive_thinking(model_name: &str) -> bool {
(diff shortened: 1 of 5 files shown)


jeremychone:
@coaoac cool. I am traveling right now, so give me few days but this looks good at first quick sight.

jeremychone:
PR commit comment: 

```
- Rename ReasoningEffort::None -> ReasoningEffort::Zero per maintainer proposal in #251;
  keyword surface stays 'none' (-none suffix keeps working), serde alias keeps old
  serialized values deserializable.
- Anthropic: Zero now short-circuits insert_anthropic_reasoning — emits
  thinking:{type:disabled} on thinking-on-by-default models (Sonnet 5, via the new
  anthropic_thinking_on_by_default helper) and omits thinking/output_config everywhere
  else (incl. Fable/Mythos, where explicit 'disabled' is rejected).
- Anthropic suffix table: -zero and -none map to Zero and are stripped; drops the
  stray 'None' (capital) => Low entry.
- Regression tests per #251 plus suffix + other-model pins (4 new tests).

Co-authored-by: Claude Fable 5 <noreply@anthropic.com>
```

jeremychone:
@coaoac Thanks for the good work. 

Couple of things. First, it is a change I will do, and second is a question. 

1) I will add a commit to normalize everything to `Zero` -zero. We will still support the -none suffix on the model name, but otherwise, we will remove the "None" emission and such. 

2) QUESTION. When Zero, why does only Sonnet 3.5 remove adaptive thinking? I would assume it would be set to false for all models that support adaptive thinking, no? 

Now, when Option::None, we should have the good default, which is what you did.

jeremychone:
@coaoac I did the 1), to be more direct about the "zero" normalization. 

Now, regarding the 2, I was not sure. Why are other models like opus 4.8 and perhaps others not default adaptive thinking true as well?

Also, I put gemini Zero, to `Budget(0)`. Provider will reject it, but at least, there is a way to set to 0 for the models that supports that.

jeremychone:
released in 0.7.0-beta.10

coaoac:
> Now, regarding the 2, I was not sure. Why are other models like opus 4.8 and perhaps others not default adaptive thinking true as well?

@jeremychone I believe only claude-sonnet-5 defaults to adaptive others don't for some reason. Fable/Mythos 5 cannot deactivate adaptive at all, would error if tried.

jeremychone:
@coaoac But for the other Opus 4.8, would it make sense, and even for models before it, to do so by default as well (genai would set it by default) when Option::None?

coaoac:
@jeremychone ah you mean if None should force adaptive on? No I think it should do nothing as you proposed and use the models native behavior.

jeremychone:
@coaoac Yes, that is what I meant. 

The reason why I sent this back was in case Opus was much better with the adaptive thinking on. 
Right now, the only way to set it is to set a reasoning level.

Anyway, they will probably make it the default in the next Opus.

coaoac:
@jeremychone did not test Opus but my guess is that they will make it default when it will make sense more often than not. Should leave it up to them or alternatively add ReasoningEffort::Adaptive. 
I would wait a bit and see what the next models do.

jeremychone:
@coaoac I like this approach. Agree

P.S. We do not need to have `ReasoningEffor::Adaptive` for now.

PR #254, opened 2026-07-06
anthropic: pass custom content parts through

ropoctl:
While we support Custom message parts in our API, might as well pass it through for Anthropic backends.


jeremychone:
released in 0.7.0-beta.10

PR #255, opened 2026-07-06
anthropic: merge extra request body fields

ropoctl:
We support extra request body in our API for other adapters, we can pass it through for Anthropic.


jeremychone:
released in 0.7.0-beta.10

PR #258, opened 2026-07-09
fix(anthropic): capture streaming cache tokens from `message_delta` fallback

Jackkakaya:
## Problem

`AnthropicStreamer::capture_usage` only reads cache tokens
(`cache_creation_input_tokens` / `cache_read_input_tokens`) from the
`message_start` event (`/message/usage/...`).

The official Anthropic API does report them there, but several
**Anthropic-compatible gateways** report them in the `message_delta` event
(`/usage/...`) instead. On those gateways, streaming responses silently return
`prompt_tokens_details.cached_tokens = 0` / `cache_creation_tokens = 0` even when
the request used `cache_control` and the server actually served/created a cache.

Concretely reproduced against **Alibaba DashScope / Qwen** Anthropic-compatible
endpoint (`.../apps/anthropic/v1`). Raw SSE from that gateway:

```
event:message_start
data:{"message":{...,"usage":{"input_tokens":3602,"output_tokens":0}}}   # no cache fields

event:message_delta
data:{"type":"message_delta","usage":{
  "cache_creation_input_tokens":0,
  "cache_read_input_tokens":3604,
  "prompt_tokens_details":{"cached_tokens":3604},
  "output_tokens":878}}                                                   # cache is here
```

Non-streaming (`into_usage`) works because it parses the final top-level `usage`
object; only the streaming path was affected.

## Fix

Fall back to `message_delta` (`/usage`) when `message_start` did not carry cache
details. `prompt_tokens` is only adjusted for `message_start` (where Anthropic's
`input_tokens` excludes cached tokens); for the `message_delta` fallback we only
surface the breakdown into `prompt_tokens_details` to avoid double counting under
gateway-specific accounting.

- Standard Anthropic: unchanged (still reads `message_start`).
- Compatible gateways reporting in `message_delta`: `cached_tokens` /
  `cache_creation_tokens` are now populated.

## Verification

Against the DashScope/Qwen Anthropic-compatible endpoint with a >1k-token
`cache_control` system prompt, streaming now reports
`prompt_tokens_details.cached_tokens = 2404` (was `0`), matching the raw SSE.
`cargo check` passes on `0.7.0-beta.11-WIP`.

### DIFF
--- a/src/adapter/adapters/anthropic/streamer.rs
+++ b/src/adapter/adapters/anthropic/streamer.rs
@@ -311,17 +311,36 @@ impl AnthropicStreamer {
 				*val += output_tokens;
 			}
 
-			// -- Capture cache tokens (only present in message_start)
+			// -- Capture cache tokens.
+			// Standard Anthropic reports them in `message_start` (`/message/usage`).
+			// Some Anthropic-compatible gateways (e.g. Alibaba DashScope / Qwen) instead report
+			// them in `message_delta` (`/usage`), so fall back to that location when `message_start`
+			// did not carry them.
 			// NOTE: Anthropic's input_tokens does NOT include cached tokens, so we must add them.
 			// See also: AnthropicAdapter::into_usage() for non-streaming equivalent.
-			if message_type == "message_start" {
-				let cache_creation: i32 = data.x_get("/message/usage/cache_creation_input_tokens").unwrap_or(0);
-				let cache_read: i32 = data.x_get("/message/usage/cache_read_input_tokens").unwrap_or(0);
+			let cache_base = match message_type {
+				"message_start" => Some("/message/usage"),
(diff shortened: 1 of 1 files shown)


jeremychone:
PR commit message

```
Anthropic reports cache_creation/read tokens in message_start, but some
Anthropic-compatible gateways (e.g. Alibaba DashScope / Qwen) report them in
message_delta instead. The streaming usage capture only looked at message_start,
so cached_tokens/cache_creation_tokens were silently 0 on those gateways.

Fall back to /usage (message_delta) when message_start did not carry cache
details. prompt_tokens is only adjusted for message_start to avoid double
counting under gateway-specific accounting.

Co-authored-by: jiangsumin <suminjackcode@gmail.com>
```

jeremychone:
@Jackkakaya thanks for this good fix. Released in 0.7.0-beta.11

PR #260, opened 2026-07-10
Add GPT-5.6 explicit prompt caching

jeremychone:
# Add GPT-5.6 explicit prompt caching

## Summary

Implement the newer OpenAI prompt-caching policy for GPT-5.6 and later model families while preserving existing behavior for older models and unrelated OpenAI-compatible providers.

## Changes

- Added GPT-5.6 and later model-family detection without relying on lexicographical version comparisons.
- Added OpenAI-specific cache-policy handling for the actual `OpenAI` and `OpenAIResp` adapters.
- Added `prompt_cache_options.mode` serialization for supported newer OpenAI requests.
- Used explicit mode without breakpoints when no cache configuration is provided.
- Used implicit mode when only a prompt cache key or request-level cache intent is provided.
- Used explicit mode for message-level cache control.
- Applied at most one breakpoint to the last eligible content block for each controlled message.
- Added protocol-specific breakpoint support for:
  - Chat Completions text, image, audio, file, and refusal blocks.
  - Responses input text, input image, and input file blocks.
- Ignored tool-level cache controls for OpenAI because the supported protocols do not provide a compatible tool-definition breakpoint representation.
- Preserved legacy `prompt_cache_retention` behavior for older OpenAI models.
- Prevented OpenAI-specific prompt-cache fields from being added to OpenAI-compatible adapters.
- Preserved normalized cache usage handling for cached tokens and cache-write tokens.
- Added regression coverage for policy selection, model detection, payload serialization, breakpoint placement, compatibility behavior, and usage normalization.

## Behavior

- GPT-5.6 and later requests without cache configuration use explicit mode without breakpoints.
- GPT-5.6 and later requests with only a prompt cache key or request-level cache control use implicit mode without an invented breakpoint.
- GPT-5.6 and later requests with message-level cache control use explicit mode with protocol-specific breakpoints.
- Tool-level cache control is ignored by OpenAI request serialization.
- Older OpenAI models retain their existing legacy cache behavior.
- OpenAI-compatible providers retain their existing shared request behavior.

## Scope

This ticket does not change the public normalized usage model or the provider-specific Anthropic cache behavior.


ropoctl:
This caused breakages for me

jeremychone:
@ropoctl Oh, how so? 

The problem is that initially, although they might have changed that, they defaulted to caching with the surcharge, which is a big no-no even for OpenAI. We cannot have a 25% surcharge as an output. 

So, making it explicit was the way I found to turn off the default caching. It has been working great for me. 

BTW, note that now, when you set breakpoints, if anything before the breakpoint has changed, it won't be part of the cache. 

Anyway, how does it break for you?

Issue #261, opened 2026-07-14
Suggest reorder request fields for better OpenAI Prompt Cache utilization


J-F-Liu:
OpenAI has confirmed that the serialized JSON request body order can affect Prompt Cache utilization. To maximize cache hits, stable fields (such as tools) should appear before frequently changing fields (such as messages).

Currently, ChatRequest derives Serialize, so the JSON field order is fixed by the struct definition. This results in requests being serialized roughly as:
```json
{
  "messages": [...],
  "model": ...,
  "stream_options": ...,
  "tools": [...],
}
```
For OpenAI, a more cache-friendly ordering would be:
```json
{
  "model": ...,
  "stream_options": ...,
  "tools": [...],
  "messages": [...]
}
```
This does not change the request semantics, since JSON object member order is not significant, but it can improve Prompt Cache utilization according to OpenAI.

Suggested solutions

Any of the following would work:

- Reorder the fields in ChatRequest (simple, but changes serialization order for all providers).
- Implement a custom Serialize for ChatRequest to control field order.

I think the third option would provide the best balance between optimization and backward compatibility.

Reference

OpenAI engineer confirmation: https://github.com/openai/openai-java/issues/316

jeremychone:
@J-F-Liu I think this was fixed in your #262 (feel free to reopen if needed)

PR #262, opened 2026-07-14
Move messages after tools in json payload for better Prompt Cache utilization

J-F-Liu:
Fix issue #261


jeremychone:
@J-F-Liu Thanks for this PR.

It is kind of crazy that the order of properties in the JSON payload matters. Most JSON libraries in most languages do not support that, at least by default.

It is good that GenAI has used custom serialization from the start.

I’ll take a look and probably do a beta release with this one.

Issue #256, opened 2026-07-06
api discussion - json-schema sanitization


ropoctl:
The various adapters all take different json schema sanitization - never quite accepting the json schema of the sorts schemars, pydantic etc generate. I implement various sanitizers on a project-by-project basis.

I could see this landing in either:
- rust-genai crate, with a chat request options bool, default false, to pass the tool json schema or output response schema to an adapter-specific sanitizer.
- a second crate, genai-utils or or other, living in this repo or another, that has this functionality.

A pro in favor of option 1- google genai py package maintains this for gemini models, and it seems like a pretty stable contract they offer that we can copy. Though I'm not sure about ant or openai models.

A pro in favor of option 2- we can avoid a lot of scope bloat in this main crate, and focus purely on llm calling and response parsing.

jeremychone:
@ropoctl First, this is a good topic to discuss and implement.

Whether it should be in genai or an auxiliary crate is also a good discussion to have. But for now, let's focus on this in the crate, and if it becomes relatively obvious that it is another scope altogether, we can split it.

Now, at a high level, I see a few options.

Option A) Flags to tell genai to sanitize the tool schema, and we have a sanitizer scheme. If it is per provider, we can add something to the Adapter interface. 

Option B) We have a Rust struct structure, which is normalized, and then we have a serializer. In the ChatRequest / Tool we keep it as a serde Value, and this will just be part of an into or try into schema.

What do you think? 

Btw, quick question, in your project, do you have those sanitizers per provider?

ropoctl:
I put a prototype in my python binding https://github.com/ropoctl/genai-pyo3/commit/bc9d68ee390b9ca9f734cc5df2e9f514f57b7af5

This API for python could be ported over directly to rust-genai PyChatOptions has the added field that the python binding crate has special logic for, but we could include this bool in rust-genai ChatOptions.

In my rust projects, I generally just have an openai sanitizer configured.

jeremychone:
@ropoctl Okay, the code you have is for your pyo3 genai. However, just to make sure I understand, in genai, this would not be related to Python, right?

ropoctl:
I am trying to keep genai-pyo3 as isomorphic binding of genai as possible. The schema sanitization rust code could be easily be lifted into this repo.

jeremychone:
Makes sense, but to be crystal clear, if we move this logic to `genai` create, it will not be related to py at all. I just want to make sure I understand this option.

[...]

PR #263, opened 2026-07-14
feat(schema): sanitize JSON Schema per adapter

ropoctl:
Refs #256.

## What

Make provider JSON Schema sanitization an adapter invariant, with separate policies for OpenAI response schemas, OpenAI strict-tool parameters, and Anthropic structured schemas.

- OpenAI Chat Completions and Responses share the same response-schema and strict-tool sanitizers.
- Anthropic uses its own structured-schema policy for JSON outputs and strict tools.
- Non-strict tool schemas are forwarded unchanged.
- Gemini remains unchanged.
- Replace the older blanket `additionalProperties` walker, which could rewrite JSON values that were not schemas.

## Minimal provider rewrites

For OpenAI structured responses and strict tools, the sanitizer only performs rewrites required by the provider or known to preserve validation semantics:

- add `additionalProperties: false` to object schemas when it is omitted;
- require every declared property, as required by OpenAI strict schemas;
- convert `oneOf` to `anyOf` only when every branch requires the same tag property with a distinct `const` value, proving that the branches cannot overlap;
- preserve `discriminator` and accepted annotations/constraints, including `title`, `description`, `default`, `nullable`, `minLength`, `pattern`, `format`, and `minimum`;
- preserve bare `$ref` uses;
- losslessly inline `$ref` uses with accepted annotation siblings because live OpenAI validation rejects every tested `$ref` sibling;
- for strict tools only, inline a top-level generated `$ref` wrapper when it resolves to an object, so the tool has the explicit object root OpenAI requires.

Anthropic uses the missing-`additionalProperties` and provably equivalent union rewrites, but does not apply OpenAI's required-property or ref-inlining rules. Optional properties, defaults, refs, and other schema semantics remain intact.

Unsupported shapes without an equivalent provider-compatible rewrite are preserved for authoritative backend validation. This includes overlapping or unproven `oneOf` branches, schema-valued dynamic maps, root arrays, and unresolved recursive pointers. There is no prompt fallback or silent weakening of constrained decoding.

## Generated-schema regressions

The unit tests contain exact raw-to-sanitized fixtures whose raw forms failed live provider validation:

- 7 structured-response cases generated by Pydantic 2.12.4, schemars 1.2.1, and Zod 3.25.76 with zod-to-json-schema 3.25.1;
- 3 strict-tool cases covering the same three generator families;
- generated ordinary objects, discriminated unions, recursive schemas, and top-level `$ref` wrappers;
- explicit regressions proving that overlapping discriminator-only branches remain `oneOf`, defaults/descriptions are preserved, bare refs remain refs, and non-strict tools remain untouched.

Live OpenAI-compatible Responses probes confirmed that the challenging raw schemas fail at baseline and their sanitized response/tool forms are accepted. Cases with no semantics-preserving rewrite continue to return native backend schema errors.

## Checks

- exact generated-fixture and adapter unit tests pass;
- `cargo fmt --all -- --check`;
- `cargo clippy --all-targets -- -D warnings`;
- Yakbak replay CI.


jeremychone:
@ropoctl Feel free to let me know when it's ready for me to look at it. I did a first scan early this morning, but I saw some changes.

So, I'll wait a little.

Let me know.

ropoctl:
I think it's pretty much ready to review

jeremychone:
@ropoctl Did a pass with pro@coder as reviewer, and I got this: 

```
Does rewriting `oneOf` to `anyOf` based only on `discriminator` weaken the schema contract? Since `discriminator` is annotation-only, overlapping branches could match simultaneously, which `oneOf` rejects but `anyOf` accepts.

Would it make sense to convert only when mutual exclusivity is proven, such as distinct required `const` tags? Otherwise, could we preserve `oneOf` and let the provider report unsupported schemas?

Would a regression test confirming that overlapping discriminator-only branches remain `oneOf` also make sense?
```

Any good points there ?

ropoctl:
> Any good points there ?

Yes, that was a good catch. discriminator alone is annotation-only and does not prove that the branches are mutually exclusive. I changed the sanitizer so oneOf is converted to anyOf only when every branch requires the same tag property with a distinct const value, including through local refs. Otherwise it preserves oneOf and lets the provider reject it. I also added regressions showing that overlapping discriminator-only branches remain oneOf. While tightening this, I live-tested the other rewrites and now preserve defaults, descriptions, constraints, discriminators, and bare refs; refs are only inlined where OpenAI rejects the sibling/root shape. The PR is ready for another look.

jeremychone:
@ropoctl Awesome!!! Will do another pass.

jeremychone:
v0.7.0-beta.14

PR #264, opened 2026-07-16
Add error handling for tool-call input JSON parsing in AnthropicStreamer, avoid stream crash

J-F-Liu:
In case of claude generate invalid json, I actually encountered this.


jeremychone:
@J-F-Liu thanks!

jeremychone:
v0.7.0-beta.14

PR #268, opened 2026-07-28
fix(anthropic): correct reasoning routing for current model names

coaoac:
Three misroutings in the Anthropic adapter's model-name predicates. They share a
cause — the predicates were written against an older naming shape and an older
model set — so they are here together rather than as three PRs. Each is
independently revertable.

## 1. `claude-opus-5` gets the legacy payload (400s)

The adapter sends:

```json
{"thinking": {"type": "enabled", "budget_tokens": 24000}}
```

`budget_tokens` was removed on that model, so the request fails.

The cause is the version comparison in `is_opus_4_7_or_higher`:

```rust
regex::Regex::new(r"claude-opus-(\d+)-(\d+)")
```

It requires two numeric segments. Current model names drop the minor version, so
`claude-opus-5` never matches — nor would `claude-opus-6`. `SUPPORT_EFFORT_MODELS`
does not list it either, so `supports_anthropic_effort` returns `false` and the
request falls through to the legacy branch.

Because `supports_anthropic_effort`, `supports_anthropic_reasoning_max`,
`supports_anthropic_reasoning_xhigh` and `supports_anthropic_adaptive_thinking`
all route through that same comparison, `claude-opus-5` additionally loses `max`,
`xhigh`, and the adaptive-thinking payload.

**Fix:** make the minor segment optional, reading a missing one as `0`. So
`claude-opus-5` is `(5, 0)` while `claude-opus-4-0` stays `(4, 0)` and remains
below the threshold. That restores the forward-compatibility the comparison was
written for — a future `claude-opus-6` works on release rather than after a table
patch.

A second, smaller change belongs to the same model: `claude-opus-5` is added to
`anthropic_thinking_on_by_default`, so `ReasoningEffort::Zero` sends the explicit
`{"type": "disabled"}`. Thinking is on by default there, so omitting the field
leaves it running. This one cannot be folded into the version comparison — Opus
4.7 and 4.8 are thinking-off by default while Opus 5 is on, so the distinction
does not follow the version ordering.

## 2. A date stamp is read as a minor version

`claude-opus-4-20250514` — the dated form of `claude-opus-4-0` — parses as version
**4.20250514** and is promoted to the modern effort path. The regex is unanchored,
so the date is captured as the minor segment. This one predates the PR; making the
minor optional does not introduce it and does not fix it.

**Fix:** a minor version is one or two digits; a longer run is a date, and the
version reads as `<major>.0`. Lookaround would express this in the pattern, but
the `regex` crate has none, so the check is in code. This also keeps a future
dated modern name correct — `claude-opus-5-20260101` reads as 5.0 and stays on the
effort path.

## 3. `xhigh` on Fable/Mythos silently becomes `max`

`supports_anthropic_reasoning_xhigh` omits Fable/Mythos, so `ReasoningEffort::XHigh`
on `claude-fable-5` falls through to the `Max | XHigh if support_reasoning_max` arm
and sends `"max"` — a more expensive level than the caller asked for, with no error.

Fable/Mythos are already listed in every sibling predicate
(`supports_anthropic_effort`, `_reasoning_max`, `_adaptive_thinking`); `xhigh` was
the only one missing them, which is what suggests an oversight rather than intent.

**Fix:** add them.

## Tests

Six, covering both directions:

- `claude-opus-5` gets `output_config.effort` and adaptive thinking
- `claude-opus-6`, `claude-opus-5-1` and `claude-opus-5-20260101` get the modern shape
- `claude-opus-4-0`, `claude-opus-4-1`, `claude-opus-4-20250514` and
  `claude-sonnet-4-5` still get `thinking.budget_tokens` — this is the guard
  against the optional minor over-promoting pre-4.7 names
- `Zero` on `claude-opus-5` sends `{"type": "disabled"}`
- `xhigh` on `claude-fable-5` and `claude-mythos-5` stays `"xhigh"`

Each fix was verified by reverting it alone and confirming the matching test goes
red — including the two additions, where the failures are
`{"type":"adaptive"}` vs `budget_tokens` for the dated name and `"max"` vs
`"xhigh"` for Fable.

`cargo test --lib` (163 passing), `cargo clippy --all-targets -- -D warnings`, and
`cargo fmt --check` are all clean.

I left `CHANGELOG.md` alone, since the history suggests you add those entries
yourself after merging. Happy to split any of the three out if you would rather
take them separately.

### DIFF
--- a/src/adapter/adapters/anthropic/adapter_shared.rs
+++ b/src/adapter/adapters/anthropic/adapter_shared.rs
@@ -850,7 +850,7 @@ fn supports_anthropic_reasoning_max(model_name: &str) -> bool {
 }
 
 fn supports_anthropic_reasoning_xhigh(model_name: &str) -> bool {
-	is_opus_4_7_or_higher(model_name) || model_name.contains("claude-sonnet-5")
+	is_opus_4_7_or_higher(model_name) || is_fable_or_mythos(model_name) || model_name.contains("claude-sonnet-5")
 }
 
 /// Models where Anthropic enables adaptive thinking by default when the request omits the
@@ -885,6 +885,12 @@ fn supports_anthropic_adaptive_thinking(model_name: &str) -> bool {
 /// and a missing one reads as `0`. Any parse or regex failure is treated as a
 /// conservative `false`.
 fn is_opus_4_7_or_higher(model_name: &str) -> bool {
+	/// A minor version is one or two digits; a longer run is a date stamp, not a
+	/// minor. Without this, `claude-opus-4-20250514` (Opus 4.0) would read as
+	/// version 4.20250514 and be promoted. Lookaround would express it in the
+	/// pattern, but the `regex` crate has none.
+	const MAX_MINOR_DIGITS: usize = 2;
+
 	static RE: OnceLock<Option<regex::Regex>> = OnceLock::new();
 	let re = RE.get_or_init(|| regex::Regex::new(r"claude-opus-(\d+)(?:-(\d+))?").ok());
 	let Some(re) = re.as_ref() else {
@@ -894,7 +900,11 @@ fn is_opus_4_7_or_higher(model_name: &str) -> bool {
 		return false;
 	};
 	let major = caps.get(1).and_then(|m| m.as_str().parse::<u32>().ok());
-	let minor = caps.get(2).map_or(Some(0), |m| m.as_str().parse::<u32>().ok());
+	let minor = match caps.get(2) {
+		Some(m) if m.as_str().len() > MAX_MINOR_DIGITS => Some(0),
+		Some(m) => m.as_str().parse::<u32>().ok(),
+		None => Some(0),
+	};
 	match (major, minor) {
 		(Some(major), Some(minor)) => (major, minor) >= (4, 7),
 		_ => false,
@@ -1368,7 +1378,7 @@ mod tests {
 	/// A future Opus release must work without a table update, in either naming shape.
 	#[test]
 	fn test_future_opus_generation_uses_effort() {
-		for model in ["claude-opus-6", "claude-opus-5-1"] {
+		for model in ["claude-opus-6", "claude-opus-5-1", "claude-opus-5-20260101"] {
 			let chat_options = ChatOptions::default().with_reasoning_effort(ReasoningEffort::Medium);
 			let options_set = ChatOptionsSet::default().with_chat_options(Some(&chat_options));
 			let target = ServiceTarget {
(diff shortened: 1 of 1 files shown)


jeremychone:
PR commit message

```
* fix(anthropic): route claude-opus-5 to output_config.effort

`is_opus_4_7_or_higher` matched `claude-opus-(\d+)-(\d+)`, requiring two
numeric segments. Current model names drop the minor version, so
`claude-opus-5` never matched and fell through to the legacy branch,
emitting `thinking: {"type": "enabled", "budget_tokens": N}` — a
parameter Anthropic removed on that model, so the request 400s.

Make the minor segment optional, reading a missing one as 0. That keeps
every pre-4.7 name below the threshold (`claude-opus-4-0` -> (4, 0)) and
restores the forward-compatibility the comparison was written for, so a
future `claude-opus-6` works on release rather than after a table patch.

Also add `claude-opus-5` to `anthropic_thinking_on_by_default`: thinking
is on by default there, so `ReasoningEffort::Zero` must send the explicit
`{"type": "disabled"}` rather than omitting the field. This one cannot be
derived from the version comparison — Opus 4.7 and 4.8 are off by
default while Opus 5 is on.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

* fix(anthropic): don't read a date stamp as a minor version; Fable/Mythos support xhigh

Two more misroutings in the same predicate family, both pre-existing.

`claude-opus-4-20250514` (Opus 4.0, the dated form of `claude-opus-4-0`)
parsed as version 4.20250514 and was promoted to the modern effort path. The
regex is unanchored, so the date is captured as the minor segment. A minor is
one or two digits; a longer run is a date, and the version reads as
`<major>.0`. Lookaround would express this in the pattern, but the `regex`
crate has none, so the check is in code.

This also keeps a future dated modern name correct: `claude-opus-5-20260101`
reads as 5.0 and stays on the effort path.

`supports_anthropic_reasoning_xhigh` omitted Fable/Mythos, so `XHigh` on
`claude-fable-5` fell through to the `Max | XHigh if support_reasoning_max`
arm and sent `"max"` — a more expensive level than the caller asked for, with
no error. Fable/Mythos are already listed in every sibling predicate
(`supports_anthropic_effort`, `_reasoning_max`, `_adaptive_thinking`); xhigh
was the only one missing them.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

---------

Co-authored-by: Amine Chaouki <amine@allergomat.se>
Co-authored-by: Claude Opus 5 <noreply@anthropic.com>
```

jeremychone:
released in 0.7.0-beta.15

PR #271, opened 2026-07-29
feat: map Anthropic ping events to ChatStreamEvent::Heartbeat (#252)

### LINKED ISSUE #252: Expose Anthropic ping events
## Summary

Anthropic `ping` SSE events are currently consumed inside `AnthropicStreamer`:

```rust
"ping" => continue,
```

That means users of `Client::exec_chat_stream` / `ChatStream` never observe provider heartbeats.

With frontier models, there may be long thinking periods, sometimes over 10 minutes. Callers of `genai` cannot distinguish between the Anthropic stream being alive and sending `ping` events, and the stream having stalled.

## Related Issues

This seems closely related to https://github.com/jeremychone/rust-genai/issues/157 and https://github.com/jeremychone/rust-genai/pull/245.

That issue/PR pair handled the control-plane `event: error`, which previously hid stream errors from callers. `event: ping` has a similar shape: it is not assistant content, but it is still operationally meaningful stream information.

## Current Behavior

The Anthropic API may send `ping` SSE events, but the Anthropic adapter discards them before they reach the public stream.

The public stream only yields semantic events, so a caller waiting on `stream.next().await` sees no difference between a series of pings and no bytes arriving at all.

## Expected Behavior

For Anthropic streams, `event: ping` should produce a non-content heartbeat event observable by the caller.

Heartbeat events should not affect captured content, captured reasoning, tool calls, usage, or stop reason.

## Proposed API

One option is to add a provider-neutral `Heartbeat` event:

```rust
pub enum InterStreamEvent {
    Start,
    Heartbeat,
    Chunk(String),
    // ...
}
```

```rust
pub enum ChatStreamEvent {
    Start,
    Heartbeat,
    Chunk(StreamChunk),
    // ...
}
```

Then Anthropic pings could map to:

```rust
"ping" => {
    return Poll::Ready(Some(Ok(InterStreamEvent::Heartbeat)));
}
```

## Alternatives

A lower-level callback such as `on_provider_event` or exposing raw SSE events would also solve this issue, but would expose provider-specific details. A provider-neutral `Heartbeat` event keeps the current `ChatStream` abstraction intact while preserving the key liveness signal needed by long-running clients.

This issue is specifically about Anthropic `event: ping` messages. The generic SSE layer currently skips bare comment keep-alives, such as lines beginning with `:`, before adapters can observe them. A provider that only sends comment keep-alives would need a separate change in the `EventSourceStream` layer.

## Suggested Tests

Add an Anthropic streaming regression fixture with `event: ping` and assert that:

- a heartbeat event is yielded to the caller
- captured content is unchanged
- captured reasoning is unchanged
- tool calls are unchanged
- usage and stop reason are unchanged

## Actual Behavior

Anthropic `ping` events are consumed internally and are invisible to the caller.

blocksdevpro:
## Summary

Closes #252

Expose Anthropic SSE `event: ping` as a provider-neutral `ChatStreamEvent::Heartbeat` so callers can tell a long-running stream is still alive (e.g. during extended thinking) instead of treating silence as a stall.

### Changes

- Add `InterStreamEvent::Heartbeat` and `ChatStreamEvent::Heartbeat`
- Anthropic streamer: map `"ping"` → `Heartbeat` (was previously discarded with `continue`)
- Keep the design adapter-agnostic so other providers can emit `Heartbeat` later if they add a similar signal
- Add yakbak cassette `tests/data/yakbak/anthropic/ping_stream/` and regression test

Heartbeats do not affect content, reasoning, tool calls, usage, or stop reason.

## Notes

- Focused single-feature PR, single commit (as requested)
- Draft PR for initial review

cc @lambdabetaeta for review if you have time

## Test plan

- [x] `cargo test --test tests_yakbak_anthropic`
- [x] `cargo test --lib`
- [ ] Maintainer review

### DIFF
--- a/examples/c21-tooluse-streaming.rs
+++ b/examples/c21-tooluse-streaming.rs
@@ -118,6 +118,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 					}
 				}
 			}
+			ChatStreamEvent::Heartbeat => {
+				println!("  Heartbeat");
+			}
 		}
 	}
 
--- a/src/adapter/adapters/anthropic/streamer.rs
+++ b/src/adapter/adapters/anthropic/streamer.rs
@@ -244,7 +244,10 @@ impl futures::Stream for AnthropicStreamer {
 							return Poll::Ready(Some(Ok(InterStreamEvent::End(inter_stream_end))));
 						}
 
-						"ping" => continue, // Loop to the next event
+						"ping" => {
+							// Map ping events to Heartbeat events to indicate the stream is still active
+							return Poll::Ready(Some(Ok(InterStreamEvent::Heartbeat)));
+						}
 						"error" => {
 							// Anthropic may emit an `event: error` mid-stream (e.g. overloaded_error,
 							// rate_limit, internal_server_error). Propagate it as a typed error so the
--- a/src/adapter/inter_stream.rs
+++ b/src/adapter/inter_stream.rs
@@ -40,4 +40,5 @@ pub enum InterStreamEvent {
 	ThoughtSignatureChunk(String),
 	ToolCallChunk(crate::chat::ToolCall),
 	End(InterStreamEnd),
+	Heartbeat,
 }
--- a/src/chat/chat_stream.rs
+++ b/src/chat/chat_stream.rs
@@ -80,6 +80,7 @@ impl Stream for ChatStream {
 						ChatStreamEvent::ToolCallChunk(ToolChunk { tool_call })
 					}
 					InterStreamEvent::End(inter_end) => ChatStreamEvent::End(inter_end.into()),
+					InterStreamEvent::Heartbeat => ChatStreamEvent::Heartbeat,
 				};
 
 				// -- OTel: record time-to-first-chunk on the first content chunk, and the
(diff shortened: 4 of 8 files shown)


jeremychone:
@blocksdevpro I did a review and ready to merge. Let know know if all good on your side.

blocksdevpro:
@jeremychone All good on my side — thank you so much for the review and for merging when ready!

Really appreciate the guidance and the chance to contribute. Looking forward to more!

jeremychone:
@blocksdevpro Ok, merged, it will be in the next 0.7.0 beta release,

I just did another commit 4f90a5f34b420abf9c713447ee46a7d09e467d33 where I made a minor cosmetic change to Heartbeat before End (as well as added the bookkeeping files)

Very good PR! Thanks

lambdabetaeta:
Thanks, and apologies for not engaging earlier... conference travel got in the way.

Post-merge review: looks exactly right. Two small notes:

1. Should [`docs/for-llm/api-reference-for-llm.md`](https://github.com/jeremychone/rust-genai/blob/0b04ffe/docs/for-llm/api-reference-for-llm.md) be updated to mention the new variant?
2. Should [`src/chat/printer.rs`](https://github.com/jeremychone/rust-genai/blob/0b04ffe/src/chat/printer.rs#L155) print something on the first heartbeat?

Issue #273, opened 2026-07-31
Add an opt-out for prompt_cache_options on ChatGPT/Codex Responses endpoints


lambdabetaeta:
Starting with `0.7.0-beta.12`, `OpenAIResp` adds `prompt_cache_options` to requests for model names recognized as GPT-5.6 or later, even when the caller has not configured prompt caching.

In the no-intent case, `openai_prompt_cache_policy` selects `OpenAiPromptCacheMode::Explicit`, and `openai_resp/adapter_impl.rs` consequently serializes:

```json
"prompt_cache_options": {
  "mode": "explicit"
}
```

This breaks Responses-compatible endpoints that do not accept this caching parameter. In particular, the ChatGPT/Codex OAuth endpoint:

```text
https://chatgpt.com/backend-api/codex/responses
```

rejects the request. The error reported by rust-genai's streaming path is:

```text
Web stream error for model 'gpt-5.6-sol (adapter: OpenAIResp)'.
Cause: HTTP error.
Status: 400 Bad Request Bad Request
Body: {"detail":"Unsupported parameter: prompt_cache_options"}
```

The request succeeds when `prompt_cache_options` is omitted. Removing `prompt_cache_key` does not suppress the field; it only changes the selected policy from implicit to explicit.

An explicit `ChatOptions` opt-out or endpoint capability setting would allow clients using ChatGPT OAuth and other compatible gateways to suppress `prompt_cache_options` while preserving the current cache-opt-in behavior for endpoints that support it.

Returning `None` unconditionally in the no-intent case may not be ideal because omission can enable the official API's default implicit caching behavior. A dedicated serialization opt-out would avoid changing those semantics.

This behavior was introduced in `0.7.0-beta.12` and remains present in `0.7.0-beta.15`.

I'm happy to put together a PR.

jeremychone:
Hi, thanks for reporting this and for offering to prepare a PR.

We want to preserve the GPT-5.6 explicit-cache default, because it prevents implicit prompt caching, which now adds cost. So, surcharges (cost or bytes) should be opt-in, not opt-out.  
However, `prompt_cache_options` must not be sent to Responses-compatible endpoints that reject it.

So, PR #274 addresses this by sending the field only to the official OpenAI Responses endpoint, while omitting it for ChatGPT/Codex and other non-allowlisted endpoints. 
It keeps caching opt-in behavior on supported OpenAI endpoints without adding a new public option.

[...]

PR #274, opened 2026-07-31
OpenAI Explicit cache only for openai.com endpoint (issue 273)

jeremychone:
# OpenAI Explicit cache only for openai.com endpoint

This respond to the #273 issue (which is related to a previous PR #260 about the new openai caching cost)

### Problem

For GPT-5.6 and later, this library sends `prompt_cache_options: {"mode": "explicit"}` when callers have not expressed cache intent. 

OpenAI's new default on 5.6 implicit caching can add cache-related cost, so the library uses explicit mode to turn off automatic caching by default. This ensures callers must deliberately opt into caching, either through general cache intent or message-level cache control, before accepting its cost and behavior tradeoffs.

However, other providers, ChatGPT/Codex Responses endpoints do not support this OpenAI-specific parameter. Requests to endpoints such as `https://chatgpt.com/backend-api/codex/responses` fail with `Unsupported parameter: prompt_cache_options`, even when the caller has not configured prompt caching.

The issue is endpoint compatibility, not the explicit-cache strategy itself. The fix preserves the cost safeguard for the official OpenAI Responses API while omitting the unsupported field elsewhere.

## Solution

- Restrict `prompt_cache_options` serialization in the Responses adapter to the official `api.openai.com` origin.
- Omit the unsupported field for ChatGPT/Codex and other non-allowlisted Responses endpoints.
- Preserve explicit mode for supported endpoints when callers have no cache intent or use message-level cache control.
- Preserve OpenAI's default implicit caching when callers provide a cache key or request-level cache control.
- Keep Chat Completions behavior unchanged.

### What changed

- GPT-5.6 Responses requests send `prompt_cache_options` only to `api.openai.com`.
- Cache key or request-level cache control omits the field, preserving API-default implicit caching.
- No-cache-intent and per-message cache control use explicit mode on supported endpoints.

### Risk

- Compatible custom Responses gateways no longer receive explicit cache options unless added to the allowlist.
- API-default implicit caching may differ from the previously explicit request behavior.
- Chat Completions behavior is unchanged.

### DIFF
--- a/src/adapter/adapters/openai/adapter_shared.rs
+++ b/src/adapter/adapters/openai/adapter_shared.rs
@@ -1,7 +1,7 @@
 //! This is support implementation of the OpenAI Adapter which can also be called by other OpenAI Adapter Variants
 
 use super::cache_policy::{
-	OpenAiPromptCacheMode, OpenAiPromptCachePolicy, OpenAiProtocol, is_gpt_5_6_or_later, openai_prompt_cache_policy,
+	OpenAiPromptCachePolicy, OpenAiProtocol, is_gpt_5_6_or_later, openai_prompt_cache_policy,
 };
 use super::schema::{OpenAiResponseFormatPlan, response_format_plan, tool_parameters_schema};
 use crate::adapter::adapters::openai::OpenAIAdapter;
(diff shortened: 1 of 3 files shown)


jeremychone:
Oo, was released in 0.7.0-beta.16, @lambdabetaeta @ropoctl let me know if this fix your issue.

PR #275, opened 2026-08-01
anthropic: preserve signed thinking across tool turns

ropoctl:
## The bug

`AnthropicStreamer` emits `ThoughtSignatureChunk` as it parses, then drops every signature on the floor at stream end:

```rust
// src/adapter/adapters/anthropic/streamer.rs
captured_thought_signatures: None,
```

Gemini and openai_resp both do `self.captured_data.thought_signatures.take()` in the same spot. So `StreamEnd::into_assistant_message_for_tool_use()`, which already exists for exactly this handoff, returns an assistant message with reasoning content and no signatures, and Anthropic rejects the continuation. Extended thinking + tool use + streaming can't round-trip.

The other half is flattening: multiple thinking blocks collapse into one reasoning string while several signatures are kept, which loses the reasoning-to-signature association needed to rebuild the blocks.

## Change

- capture signatures at stream end, same as the other two adapters
- keep each reasoning/signature pair in block order instead of flattening
- same pairing on the non-streaming path
- rebuild signed thinking blocks ahead of text/tool_use when serializing an assistant turn
- mirror signatures onto the first tool call as a convenience, not as the canonical source

Signature assembly is a plain append: `content_block_start.content_block.signature` seeds the buffer, `signature_delta` chunks concatenate onto it. Signatures are opaque base64, so nothing tries to detect resends or overlapping chunks.

If reasoning and signature counts don't line up, the thinking blocks are dropped with a warning and the rest of the message is sent. That happens with a history carried over from a provider returning unsigned reasoning, or an assistant message built by hand. Pairing by position would sign the wrong text, which Anthropic rejects anyway.

Signature capture follows `capture_reasoning_content`, since a signature signs the reasoning text and is unusable without it. Nothing is captured unless you opt in. The requirement is documented on `with_capture_reasoning_content` and `into_assistant_message_for_tool_use`, and warned once per stream when only `capture_tool_calls` is set.

## Cross-provider

This doesn't add an Anthropic-shaped concept. `captured_thought_signatures` and `ContentPart::ThoughtSignature` already exist and are already populated by Gemini and openai_resp. Anthropic was the one adapter hardcoding `None`, so this brings it in line rather than introducing a parallel mechanism.

## Tests

- recorded replay: real Anthropic stream, empty start signature plus `signature_delta`, tool call, continuation
- synthetic replay: two thinking blocks with non-empty start signatures, tool continuation, pairing asserted end to end
- recorded two-turn adjudication cassette, multi-tool round trip
- non-streaming replay plus a two-block response/request round-trip
- unit: signature chunks concatenate verbatim; unpaired metadata dropped for signature-without-reasoning, unsigned reasoning, and unequal counts

`cargo fmt --all -- --check` clean, 238 lib tests, 31 replay tests across 9 yakbak targets.

## Since the last review

- signature assembly reduced to a plain append; removed the overlap/resend detection, which could match by coincidence on base64 and corrupt a signature
- unpaired reasoning/signature metadata no longer fails the request; `Error::AnthropicThinkingBlockMismatch` is gone
- capture-flag requirement documented and warned about at runtime instead of failing later at the provider
- rebased on main

On the suggestion to capture signatures under `capture_tool_calls` alone: a signature signs the reasoning text, so capturing one without it produces a block that can't be rebuilt, and it would now be dropped as unpaired. Documenting the flag pair seemed better than allocating something unusable, but happy to revisit if you'd rather it went the other way.


jeremychone:
@ropoctl Okay, FYI, I have been reviewing this on my side.

I am trying to make sure we reuse the capture... and that these extra captures are opt-in.

Anyway, give me a day or two while I put everything together.

Also, I use AI a lot, and I am completely okay with using AI for PRs and such, but it is important to have a "real coder's tone" in the description.

Anyway, no worries, I am working on this, and I will come back with feedback.

jeremychone:
@ropoctl I have been late on this one. 

So, I have two main concerns about this one: 

1) More work should be opt-in, not opt-out. Here, it seems we are going to capture more stuff, so can we use the `capture_reason`... opt-ins? 

2) Can this fit across providers? For example, Gemini has some thinking signature as well. 


What do you think about those two points?

ropoctl:
> @ropoctl I have been late on this one.
> 
> So, I have two main concerns about this one:
> 
> 1. More work should be opt-in, not opt-out. Here, it seems we are going to capture more stuff, so can we use the `capture_reason`... opt-ins?
> 2. Can this fit across providers? For example, Gemini has some thinking signature as well.
> 
> What do you think about those two points?

i dont think this should be opt in because the request body is the same either way and the provider sends it regardless. but i think that is the precedent we have for other places here- tossing out stuff from the response body that is sent regardless when the capture_x() is not explicitly set. i somewhat disagree with this and we can discuss later, but i have revised it accordingly to respect the capture_reasoning flag.

jeremychone:
@ropoctl I can see your point about optins or not. 

I think, what I was thinking about is since this is for the "round trip," it does impact what we send back, no?

ropoctl:
> @ropoctl I can see your point about optins or not.
> 
> I think, what I was thinking about is since this is for the "round trip," it does impact what we send back, no?

not really, because the application level (user of this library) is responsible for writing the agentic loop or whatever to construct the ChatRequest for the next request. but if they dont opt into capture of tools/signatures/etc, then obviously it will be impossible for them to construct such at the app level

jeremychone:
@ropoctl I am a little worried about this one, because it seems to be Anthropic-specific. Now, I am not that deep into this flow, and your approach might be spot on.

however, here is what my AI check process gave me back:

````md

#### Cons & Quality Concerns

- **Heuristic overlap deduplication in `append_signature_delta`**: The logic in `ThinkingBlock::append_signature_delta` tries to detect overlapping string slices between sequential chunks using a heuristic suffix/prefix match loop. Because thought signatures are opaque base64-encoded crypto blobs, random character collisions can trigger false positive overlaps, corrupting the signature string and causing signature rejection by Anthropic.

- **Event queue management**: Introducing `pending_events: VecDeque<InterStreamEvent>` in `AnthropicStreamer` adds state complexity to the stream poll loop that could lead to desynchronization if not carefully managed.

### 2. Public API Impact

PR 275 introduces several changes to the public API surface:

- **New Error Variant**: Adds `Error::AnthropicThinkingBlockMismatch { reasoning_count: usize, signature_count: usize }`.
- **New Public Helper**: Adds `ChatResponse::into_assistant_message_for_tool_use(self) -> Option<ChatMessage>`.
- **Content Part Semantics**: Previously, Anthropic non-streaming responses only exposed `reasoning_content` at the top level and did not push `ContentPart::ThoughtSignature` or `ContentPart::ReasoningContent` into `response.content.parts`. With PR 275, `response.content` now contains both `ThoughtSignature` and `ReasoningContent` parts. Callers iterating over `response.content` will encounter these parts for Anthropic models where they previously only saw `Text` and `ToolCall`.

### 3. Behavioral Side Effects & Safety Issues

#### 3.1 **Hard Break on Multi-turn or Model Switching (Critical)**:
  In `into_anthropic_request_parts`, the PR enforces a strict equality check:
  ```rust
  if reasoning_contents.len() != thought_signatures.len() {
      return Err(Error::AnthropicThinkingBlockMismatch { ... });
  }
  ```
  If an application maintains a conversation history across different providers (for example, starting with DeepSeek or OpenAI where reasoning has no signature, and subsequently passing the history to Anthropic), or if a user manually sets `.with_reasoning_content("...")` on an assistant message without a signature, the request will immediately hard-fail with `AnthropicThinkingBlockMismatch`. Instead of failing the entire request, missing signatures should be handled gracefully (for instance, dropping or converting the thinking block if Anthropic cannot accept unsigned thinking).

#### 3.2 **Coupling of Signature Capture with Reasoning Capture**:
  In `AnthropicStreamer`, capturing signatures is gated strictly on `self.options.capture_reasoning_content`. If a caller enables `capture_tool_calls(true)` but leaves `capture_reasoning_content(false)`, signatures are not captured, which prevents the subsequent tool continuation request from succeeding with Anthropic.

#### 3.3 **Cross-Provider Inconsistency**:
  Other providers (such as Gemini and Bedrock) also use thought signatures or reasoning blocks. Introducing Anthropic-specific internal structures (`InterStreamThoughtBlock`) in `InterStreamEnd` creates provider asymmetry rather than a generalized representation for signed thinking blocks across providers.

### Conclusion & Recommendations

Do not merge PR 275 yet. Request the following improvements from the author:

A. Replace the heuristic string-slice overlap detection with exact delta appending or clean protocol-based delta handling.

B. Avoid hard errors on reasoning/signature mismatch; handle unsigned reasoning gracefully when constructing Anthropic requests.

C. Ensure tool-call capture can capture signatures necessary for tool continuations without requiring full text reasoning capture, or document the exact capture flags required.

D. Consider standardizing signed thinking representation across Gemini and Anthropic.
````

Anything worth addressing there?

**Comment by ropoctl** [comment]
i think the ones that make sense are addressed

**Comment by jeremychone** [comment]
let me look at it.

**Comment by jeremychone** [comment]
For the record, is the AI review of the latest update of this PR. 


```

### Review Assessment

#### 1. Code Quality

The code quality in the revised PR is solid:

- **Clean Signature Delta Assembly:** The earlier heuristic string-slice overlap detection in `append_signature_delta` was removed in favor of a direct, robust append. Because Anthropic signatures are opaque base64 strings, concatenating the seed from `content_block_start` with subsequent `signature_delta` chunks avoids any risk of false-positive overlap deduplication.
- **Robust Stream State Management:** The `AnthropicStreamer` introduces a `ThinkingBlock` accumulator and a small `pending_events: VecDeque<InterStreamEvent>` to buffer block-start events (signature chunk and reasoning chunk) so they are drained smoothly without disrupting stream polling.
- **Extensive Test Coverage:** The PR includes thorough unit tests for `ThinkingBlock` assembly, non-stream and stream serialization tests, multi-block round-trip tests, and realistic yakbak replay cassettes (including multi-tool adjudication reproductions).

#### 2. Public API Impact

The changes to the public API are backwards-compatible and additive:

- **New Method on `ChatResponse`:** Adds `ChatResponse::into_assistant_message_for_tool_use(self) -> Option<ChatMessage>`, mirroring the existing helper on `StreamEnd`.
- **Existing `StreamEnd` Helper:** `StreamEnd::into_assistant_message_for_tool_use(self) -> Option<ChatMessage>` retains its existing public signature while properly carrying forward reasoning and signatures.
- **Content Parts:** In `AnthropicAdapter::build_chat_response`, `response.content` now contains `ContentPart::ThoughtSignature` and `ContentPart::ReasoningContent` in addition to `Text` and `ToolCall` parts, matching Gemini and OpenAI Responses behavior.
- **No Breaking Error Variants:** The previously proposed `Error::AnthropicThinkingBlockMismatch` error variant has been removed.

#### 3. Behavioral Side Effects and Safety

- **Graceful Handling of Multi-Provider Histories:** If a conversation history containing unsigned reasoning (such as from OpenAI or DeepSeek) or hand-crafted reasoning is sent to Anthropic, the adapter checks whether reasoning block counts match signature counts. If they do not align, it logs a `tracing::warn!` and safely omits the thinking blocks rather than failing the request.
- **Capture Opt-in Alignment:** Signature capture is coupled with `capture_reasoning_content`, which is logical because a thought signature cryptographically signs its associated reasoning text. A runtime warning (`warned_unreplayable_thinking`) fires once per stream if `capture_tool_calls` is enabled without `capture_reasoning_content` when thinking blocks are received.
- **Internal Structs:** The introduction of `InterStreamThoughtBlock` on `InterStreamEnd` is purely crate-internal and does not leak into public types.

### Verdict

PR 275 is **safe to merge**. The prior review feedback regarding heuristic string matching and hard error breaking has been completely resolved.

```

So, all GOOD!

**Comment by jeremychone** [comment]
PR commit message: 

```
* anthropic: preserve signed thinking across tool turns

* test: add two-turn Anthropic adjudication cassette

* test: record adjudication flow with Opus 5

* ^ anthropic - concat signature deltas, drop unpaired thinking blocks

Signature deltas now append verbatim. The previous overlap/resend detection
could match by coincidence on opaque base64 and silently corrupt a signature;
the protocol only ever concatenates, seeded by content_block_start.

Unpaired reasoning/signature metadata no longer fails the request. Counts can
differ for a history carried over from a provider that returns unsigned
reasoning, or an assistant message built by hand. Pairing by position would
sign the wrong text, so the thinking blocks are dropped with a warning and the
rest of the message is sent. Removes Error::AnthropicThinkingBlockMismatch.

Signature capture stays gated on capture_reasoning_content, since a signature
signs the reasoning text and is unusable without it. That requirement is now
documented on both options and warned about once per stream when only
capture_tool_calls is set.
```

jeremychone:
released in `0.7.0-beta.21`

Issue #269, opened 2026-07-28
Anthropic adapter silently drops Text content parts on ChatRole::Tool messages


JMLX42:
## Summary

A `ContentPart::Text` attached to a `ChatRole::Tool` message is silently discarded when building an Anthropic request. This makes Anthropic's `[tool_result, text]` content shape unreachable through genai, and pushes callers onto a workaround that fails on Bedrock.

Version: genai 0.6.5.

## Where

`src/adapter/adapters/anthropic/adapter_shared.rs`, the `ChatRole::Tool` arm:

```rust
ChatRole::Tool => {
    let mut values: Vec<Value> = Vec::new();
    for part in msg.content {
        if let ContentPart::ToolResponse(tool_response) = part {
            values.push(json!({
                "type": "tool_result",
                "content": tool_response.content,
                "tool_use_id": tool_response.call_id,
            }));
        }
    }
    if !values.is_empty() {
        let values = apply_cache_control_to_parts(cache_control.as_ref(), values);
        messages.push(json!({"role": "user", "content": values}));
    }
}
```

The `if let` filters. Any non-`ToolResponse` part is dropped with no error and no warning.

## Why it matters

Anthropic has no `tool` role — a tool result is a content block inside a `user` message. That means a `user` turn can legitimately carry a tool result *and* a text block:

```json
{"role": "user", "content": [
  {"type": "tool_result", "tool_use_id": "...", "content": "..."},
  {"type": "text", "text": "..."}
]}
```

This is the shape Claude Code itself emits when it appends instruction text alongside a tool result. Through genai it cannot be produced: attaching a `ContentPart::Text` to the `ChatRole::Tool` message is the natural expression of it, and that text is silently dropped.

The available workaround is to append a separate `ChatRole::User` message after the tool message. That yields two consecutive `role: "user"` entries on the wire, because the adapter lowers `ChatRole::Tool` into a user message. Anthropic's first-party API tolerates this — it combines consecutive same-role turns automatically. **AWS Bedrock does not**: the auto-combining change shipped to the first-party API and not to Bedrock, so Bedrock still enforces strict alternation and returns

```
roles must alternate between "user" and "assistant", but found multiple "user" roles in a row
```

[...]

Issue #65, opened 2025-06-29
Support Rate Limit Error


Lommix:
Hey all,

one common error is hitting some sort of rate limit. Especially with claude. In my application, I currently do some
error string matching to figure out if I hit a rate limit, then sleep and repeat.

I suggest, to introduce a new Error variant to genai to represent a rate limit, identified by simple match on an error result with common terms like "rate_limit_error".

What do you think?

jeremychone:
@Lommix So, I like the idea, but then we would have to implement it for each provider, and even the OpenAI family of APIs will probably have differences.

This could be quite a bit of work and may not be reliable. But I can be persuaded otherwise and would accept a PR for it.

Issue #27, opened 2024-09-29
LLM request logging


Boscop:
It would be nice to be able to log all chat history easily :)

jeremychone:
This is good as well. Few thoughts

- I will probably add tracing.
- There was a related ticket about advanced remote tracing or something.
- I will also likely offer a "hook API" or some MPSC mechanism, to allow registering for those events in some way. This way, the library user can subscribe to the events and implement their desired logic.​

There might be some methods to add to chat-request to format the entire chat in a convenient way. This could be simple and add significant value.

PR #277, opened 2026-08-07
Exec hooks, streaming error headers, tool-result binary parts, Anthropic URL images, embedded tool-response fixes

VikashLoomba:
Hi @jeremychone, I'm a developer working on a project that leverages the rust genai crate. As a part of my usage of the library, I came across some issues that I thought you might appreciate a pull request for. I know lately a lot of repo's have outright banned AI written PR's. I felt these commits were genuinely helpful to the genai project, so I felt compelled to still personally write this note and put the PR up for you to review or pull implementations from. Each commit is structured so that pieces of it can be merged independently, in case there are things you don't want to upstream. 

Just as an aside, I think the rust-genai project is an important OSS project, as it fills in an important gap in the rust ecosystem, and builds the primitives that other Rust projects can benefit from, so thank you for creating it.

Everything below was AI generated:
---

Five commits adding request/response observability, richer error surfaces, and tool-result fidelity to genai. Motivation: we are building a Rust agent runtime on top of genai that needs behavioral parity with the pi ecosystem's production agent stack; these are the capabilities genai's request path could not express.

**Issue linkage**: no open issue is fully resolved by this PR, so none are marked as closed. It is *Related to #65, #27, and #269* as noted inline below. The silent data-loss bugs fixed by commits 2, 4, and 5 were previously unreported.

### 1. `!` webc — expose response headers on streaming http errors (`13fabab`)
`Error::HttpError` gains `headers: Box<HeaderMap>` (same representation the non-streaming `webc::Error::ResponseFailedStatus` already carries), populated at both streaming construction sites: the shared `WebStream` lazy-send failure path and Bedrock's byte-stream transport. Downstream retry layers need `retry-after`, `retry-after-ms`, and `x-should-retry` to honor provider-requested retry delays; on the streaming path those headers were invisible.
**Related to #65**: this is deliberately the provider-neutral primitive — raw status + headers with no per-provider interpretation — so callers can implement rate-limit detection and retry-delay honoring without the per-provider variance concern raised in that thread. A typed rate-limit variant can layer on top later if desired.
**Migration (`!`)**: exhaustive matchers/constructors of `Error::HttpError` must add the `headers` field or match with `..`.
**Note**: includes a one-line fix for the pre-existing `--features otel` compile break on main (missing `CacheBreakpointNoEligibleContent` match arm) — happy to split this into its own PR if preferred.

### 2. `!` tool — ToolResponse binary parts (`c364f3e`)
`ToolResponse` gains `parts: Option<Vec<Binary>>` (+ `with_parts`/`append_binary` builders) so tool results can carry images. Per-adapter serialization covers every provider:
- **Native blocks**: Anthropic `tool_result` content arrays, Bedrock Converse `toolResult` blocks, OpenAI Responses `function_call_output` output arrays (`input_text`/`input_image`).
- **Follow-up user-message fallback** where the tool slot is string-only: OpenAI Chat Completions (covers all OpenAI-compatible providers) and Gemini — tool slot keeps text or a `"(see attached image)"` placeholder, images ride one `"Attached image(s) from tool result:"` user message batched per tool run (positioned to preserve the required tool-message adjacency; on Gemini, emitted after the run so the functionResponse merge invariant is preserved).
- **Ollama**: native `images` array on a follow-up user message. Custom-tool outputs on Responses (raw-string wire) get the same rescue.
Text-only responses serialize byte-identically to before on every adapter (regression-pinned).
**Migration (`!`)**: `ToolResponse` struct literals must add `parts: None`, or use the builders.

### 3. `+` client — payload interceptor and response observer (`dc8621e`)
Two per-request exec hooks modeled on the `AuthResolver` idiom, installed via `ClientBuilder`: `PayloadInterceptor` (async; receives `ModelIden` + serialized payload, may replace it — `None` keeps it) and `ResponseObserver` (async; receives `ModelIden`, `StatusCode`, `HeaderMap` as soon as the response arrives, before the body is consumed, including 4xx/5xx). Covers non-streaming and all streaming transports (threaded through the crate-private adapter/webc internals; purely additive public API). Used for request auditing and attribution headers.
**Related to #27**: this is intended as the "hook API … so the library user can subscribe" direction sketched there, modeled on the existing `AuthResolver` idiom so it reads like the crate's own resolver surfaces. Both hooks are purely additive and opt-in; zero cost and byte-identical behavior when unset (regression-pinned).

### 4. `^` anthropic — URL image sources (`9b5603f`)
The Anthropic serializers emitted only base64 image sources and silently skipped `BinarySource::Url` images (with a warning), though the Messages API supports `{"type":"url"}` sources. Both the user-message path and the tool-result parts path now emit the native URL source; base64 emission is byte-identical (regression-pinned). Delegating providers reusing the Anthropic serializers inherit this.

### 5. `-` chat — embedded tool responses (`cee6008`)
Fixes silent data loss for `ContentPart::ToolResponse` appearing outside Tool-role messages, under one rule: supported shapes are translated to what each wire actually represents; unsupported shapes fail loudly; nothing vanishes silently.
- **User-embedded** (native shape on the Anthropic/Bedrock/Gemini wires): OpenAI CC, OpenAI Responses, and Ollama previously dropped or garbled these — they now extract them as standalone `role:"tool"` messages / `function_call_output` items placed before the remaining user content (preserving `tool_calls` adjacency, incl. against pending image flushes); images fold into the carrying user message where the wire allows; an emptied user message is omitted.
- **Assistant-embedded** (no provider wire represents a tool result authored by the assistant): previously silently dropped (OpenAI, Anthropic, Bedrock, Gemini) or garbled (Ollama) — now a uniform hard error using the existing `Error::MessageContentTypeNotSupported`, directing callers to Tool-role messages. No new error variant.
- Also fixes Ollama's Tool-role path losing all but the last response when one Tool message carries several `ToolResponse` parts (now one tool message per response; single-response wire output byte-identical).

**Design note on the assistant-embedded hard error**: an assistant message carrying a tool *result* has no representation on any provider wire, so the only non-silent options are an invented placement or an error; we chose the error (via the existing variant, no new API) because fabricated placements change conversation semantics behind the caller's back. Happy to soften to warn-and-skip if you prefer leniency here — the user-embedded translation is independent of that choice.
**Related to #269**: fixes the OpenAI-family user-embedded `ToolResponse` drop documented in that issue and adopts its "nothing vanishes silently" policy for embedded tool responses; the Anthropic Tool-role `Text`-part drop that is #269's primary ask is intentionally out of scope here.

## Design notes

- Every adapter change follows that adapter's existing serialization patterns; the exec hooks are threaded through the existing crate-private internals (`Adapter`/`AdapterDispatcher`/`WebStream` signatures), with no dispatch restructuring and no new public types beyond the two hook surfaces.
- Public API impact: additive except the two `!` items above, each with a one-line migration.

## Testing

- `cargo test --lib`: 250 passed (272 with `--features otel`, which also fixes the otel build on main)
- Replay CI suites all green: `tests_yakbak_openai_resp` (8), `tests_yakbak_gemini` (2), `tests_yakbak_github_copilot` (2), `tests_yakbak_ollama_cloud` (1), `tests_yakbak_aihubmix` (1)
- `cargo check --all-targets`, `cargo clippy --all-targets` (0 warnings), `cargo fmt --check` clean
- ~60 new offline unit tests pin exact wire shapes per adapter, plus byte-level regressions for previously-existing serialization
- `dev/specs/*` updated for every module touched; CHANGELOG entries in the unreleased section per house style, including migration notes for the two `!` changes


ropoctl:
this should be about 6 separate prs because its many independent issues.

jeremychone:
@VikashLoomba First, thanks for the PR and your transparent comment.

I am all for AI accelerating coding, as long as we humans micromanage the code design, features, and so on, especially for libraries at the level of `genai`.

As @ropoctl  says, this definitely needs to be split into different PRs, where each PR should usually contain one commit and focus on adding one feature or capability, or fixing one bug.

So, I am going to close this one, but feel free to submit a new PR.

Also, make sure to review what the AI does, and work with the AI to write a very concise description for each PR. Make sure everything is tightly scoped per PR.

Perhaps start by submitting the lower-level PRs or the bug fix, and then we can work our way up to higher-level or new functionality.

PR #279, opened 2026-08-10
- fix(anthropic): align stream usage capture with new anthropic api behaviour

### LINKED ISSUE #272: Anthropic streaming over counts captured `Usage` fields
## Summary

When streaming with `capture_usage(true)`, every field of the captured `Usage` is over-counted:
| field | genai reports | should be |
| --- | --- | --- |
| `prompt_tokens` |  ~2× the real value | the provider's cumulative total |
| `completion_tokens` | inflated by few tokens | the final `message_delta` value |
| `total_tokens` | sum of the two wrong numbers | correct amount |

Anthropic reports usage in `message_start` and again in each `message_delta`, and both are cumulative absolute snapshots, not increments:

> The token counts shown in the `usage` field of the `message_delta` event are **cumulative**.
> — https://platform.claude.com/docs/en/build-with-claude/streaming

`AnthropicStreamer` adds the snapshots together instead of taking the latest.

## Proof

Raw SSE from `api.anthropic.com/v1/messages`, `claude-haiku-4-5`, "write a 1500-word essay":
```
event: message_start
data: {"type":"message_start","message":{...,"usage":{"input_tokens":26,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"output_tokens":3}}}

... content_block events ...

event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"end_turn"},"usage":{"input_tokens":26,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"output_tokens":1956}}
```

The snapshot semantics are visible directly: `output_tokens` climbs 3 → 1956 (running total, not an increment), while `input_tokens` repeats the same 26. Summing the events gives `prompt_tokens = 52` and `completion_tokens = 1959`, genai reports exactly those. The correct read is 26 / 1956.

## Where

`src/adapter/adapters/anthropic/streamer.rs`, in `capture_usage` — both `message_start` and `message_delta` reach this code, and it accumulates:

```rust
if let Ok(input_tokens) = data.x_get::<i32>(input_path) {
    let val = self.captured_data.usage
        .get_or_insert(Usage::default())
        .prompt_tokens
        .get_or_insert(0);
    *val += input_tokens;      // <-- adds 26, then adds 26 again
}
// same += for output_tokens
```

## Cause

The `+=` was correct when written, back then `message_delta.usage` carried only `output_tokens`, so each field had a single source and the addition was effectively an assignment. Anthropic later added the full usage object (input + cache fields) to `message_delta`, turning the accumulation into a silent over-count. 

## Suggested fix

Treat every usage field as a snapshot: last value wins, overwriting only the fields the event actually carries. Build the final `Usage` once at `message_stop`, with the same arithmetic the non-streaming `into_usage()` already uses (`prompt = input + cache_creation + cache_read`, `total = prompt + completion`).

Happy to open a PR with this fix if the approach sounds right to you?

CristiC0:
Closes #272.

Fix for the Anthropic streaming usage over-count. `capture_usage` in `AnthropicStreamer` now treats each usage payload as a cumulative snapshot. Every event overwrites the token fields it actually carries, so the final capture holds the latest values instead of the sum of `message_start` + `message_delta`.

Also added replay tests and fixtures (recorded from the live API).

### DIFF
--- a/src/adapter/adapters/anthropic/streamer.rs
+++ b/src/adapter/adapters/anthropic/streamer.rs
@@ -283,13 +283,12 @@ impl futures::Stream for AnthropicStreamer {
 impl AnthropicStreamer {
 	fn capture_usage(&mut self, message_type: &str, message_data: &str) -> Result<()> {
 		if self.options.capture_usage {
-			let data = self.parse_message_data(message_data)?;
-			// TODO: Might want to exit early if usage is not found
+			let mut data = self.parse_message_data(message_data)?;
 
-			let (input_path, output_path) = if message_type == "message_start" {
-				("/message/usage/input_tokens", "/message/usage/output_tokens")
+			let usage_path = if message_type == "message_start" {
+				"/message/usage"
 			} else if message_type == "message_delta" {
-				("/usage/input_tokens", "/usage/output_tokens")
+				"/usage"
 			} else {
 				// TODO: Use tracing
 				tracing::debug!(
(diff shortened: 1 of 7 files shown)


jeremychone:
@CristiC0 Thank you, all merge. Will be in the next `0.7.0-beta` release. (planning to do officinal v0.7.0 sometime this month)

jeremychone:
@CristiC0 Thank you for the fix. I am not using `otel` so, really appreciate your PR.

PR #278, opened 2026-08-09
fix(otel): cover CacheBreakpointNoEligibleContent in error.type derivation

VikashLoomba:
Following up on #277 — splitting it into small, single-purpose PRs as you asked. This is the first: the smallest, lowest-level piece, a bug fix.

## Problem

The `otel` feature has not compiled since v0.7.0-beta.18. `error_type()` in `src/otel/error.rs` matches `Error` exhaustively with no wildcard arm, and the `CacheBreakpointNoEligibleContent` variant added in beta.18 was never mapped, so `cargo check --features otel` fails with E0004 (non-exhaustive patterns).

## Change

- `src/otel/error.rs`: map `Error::CacheBreakpointNoEligibleContent { .. }` → `"cache_breakpoint_no_eligible_content"`.
- `CHANGELOG.md`: one OpenTelemetry fix line.

## Testing

`cargo build`, `cargo test --lib` (198 pass), `cargo test --lib --features otel` (220 pass — previously did not compile at all), `cargo fmt --check`, `cargo clippy` clean.

Depends on nothing. The remaining pieces from #277 will follow one at a time, low-level first, once this lands.

PR #280, opened 2026-08-13
feat(webc): expose response headers on streaming HTTP errors

VikashLoomba:
Part of the #277 split (single-purpose PRs, low-level first, gated on merges). The otel-compile prerequisite (#278) is now merged, so this is a single commit on top of current `main`.

Streaming HTTP failures (`WebStream` and the Bedrock byte stream) previously discarded the response headers, so downstream retry layers could not see `retry-after`, `retry-after-ms`, or `x-should-retry` on the streaming path — unlike the non-streaming `webc::Error::ResponseFailedStatus`, which already carries a `HeaderMap`. This adds a `headers: Box<HeaderMap>` field to `Error::HttpError`, captured while the `reqwest::Response` is still in hand.

**Changed**
- `src/error.rs`: `Error::HttpError` gains `headers: Box<HeaderMap>`.
- `src/webc/web_stream.rs` + `src/adapter/adapters/bedrock/shared.rs`: capture headers before reading the error body; new lib test `test_web_stream_http_error_captures_headers` (one-shot local TCP server asserting the header round-trip).
- `src/otel/error.rs`: add `headers` to the existing `HttpError` test constructor.
- CHANGELOG / `dev/specs/spec-webc.md` / `docs/for-llm/api-reference-for-llm.md`.

**Breaking**: exhaustive matchers/constructors of `Error::HttpError` must add `headers` (or match with `..`).

**Tests**: build, `test --lib` (199), `test --lib --features otel` (221), `fmt`, `clippy` clean; the five yakbak replay suites pass (touches the bedrock adapter).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v0.7.0-beta.x (see [genai versions](https://crates.io/crates/genai/versions))
 
+- `!` API CHANGE - `Error::HttpError` adds a `headers: Box<HeaderMap>` field carrying the response headers of failed streaming HTTP calls (e.g., `retry-after`, `retry-after-ms`, `x-should-retry`), matching the non-streaming `webc::Error::ResponseFailedStatus`, so downstream retry layers can honor provider-requested retry delays. Exhaustive matchers/constructors of `Error::HttpError` must add the `headers` field (or match with `..`).
 - `!` API CHANGE - `Tool` adds the public `custom_format: Option<Value>` field for provider-native freeform custom-tool formats. Downstream `Tool` struct literals must add `custom_format: None`, or preferably migrate to `Tool::new(...)` and builder methods. `Tool::with_custom_format(...)` is the new builder API.
 - `+` New Providers:
   - AtlasCloud - default env: `ATLASCLOUD_API_KEY`, Adapter: OpenAI, endpoint: `https://api.atlascloud.ai/v1/` (activated on the `atlascloud::` namespace) (PR #259)
--- a/dev/specs/spec-webc.md
+++ b/dev/specs/spec-webc.md
@@ -31,6 +31,6 @@ The module consists of three main internal components:
 
 - **Generic JSON Response Handling:** `WebResponse` abstracts successful non-streaming responses by immediately parsing the body into `serde_json::Value`. This allows adapter modules to deserialize into their specific structures subsequently.
 
-- **Error Richness:** The `Error::ResponseFailedStatus` variant includes the `StatusCode`, full `body`, and `HeaderMap` to provide comprehensive debugging information upon API failure.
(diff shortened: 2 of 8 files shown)


jeremychone:
@VikashLoomba Thanks for this good, narrow PR. 

Code looks good, functionality looks reasonable. 

What worries me is that now we might capture some auth information, like session IDs and such. 

Not sure, might be okay. Let me think about this. 

But it's definitely something valuable to have.

VikashLoomba:
Thanks Jeremy — good instinct to check. A few things that might help the call:

**These are response headers, not request headers.** The request-side secrets we send (`Authorization`, `x-api-key`, bearer tokens) never appear here — this only captures what the server returns on a failed response. The realistic sensitive case is exactly the one you flagged: `Set-Cookie` / session IDs, which a few providers do emit.

**It matches what the non-streaming path already does.** `webc::Error::ResponseFailedStatus` has carried `headers: Box<HeaderMap>` for a while; this PR just brings the streaming path to parity, so it closes an asymmetry rather than introducing a new capture surface.

**Neither variant leaks headers via `Display`.** Both `HttpError` and `ResponseFailedStatus` format only status + body, so `{}` / `.to_string()` / normal error messages never print them. `Debug` (`{:?}`) would include the map — the same caveat that already applies to the non-streaming variant today.

If you'd rather be conservative, an easy option is to store a **filtered subset** instead of the whole map — an allowlist of the retry-relevant ones (`retry-after`, `retry-after-ms`, `x-should-retry`, `ratelimit-*`) — which fully sidesteps `Set-Cookie` while keeping the retry value. Happy to switch to that (applied uniformly to both variants) if you prefer, or keep the full map. No rush — whatever you land on, I'll adjust.

jeremychone:
@VikashLoomba Yes, I think you are right. In this case (API), it is fine because it is the response. 

The reason why I was cautious is that sometimes for web apps (usually web UIs), the server sends back a new token with an updated expiration. It is a pattern I like. 

But in this case again, with the API, this is not a thing, since they are all API key-based. 

So I think we are fine. 

Thanks for your thoughtful answer. 

Merged, will be in the next beta release (in the next couple of days.)

jeremychone:
PR commit message

```
Streaming HTTP failures (WebStream and the Bedrock byte stream) now surface
the response headers on `Error::HttpError` via a new `headers: Box<HeaderMap>`
field, captured while the `reqwest::Response` is still in hand, matching the
non-streaming `webc::Error::ResponseFailedStatus`. Motivation: downstream
retry layers need `retry-after`, `retry-after-ms`, and `x-should-retry` to
implement provider-requested retry delays; previously those headers were
invisible on the streaming path.

Exhaustive matchers/constructors of `Error::HttpError` must add the `headers`
field (or match with `..`).
```

PR #281, opened 2026-08-14
feat(tool): ToolResponse binary parts (images in tool results across adapters)

VikashLoomba:
Part of the #277 split (single-purpose PRs, low-level first, gated on merges). Its only prerequisite — the otel compile gate (#278) — is merged, so this sits as a single commit on current `main`.

`ToolResponse` was text-only, forcing agentic harnesses to degrade tool-produced images (screenshots, file reads) to text markers even though provider APIs can carry them. This adds an additive `parts: Option<Vec<Binary>>` field (serde-skipped when absent) with `with_parts` / `append_binary` builders, and serializes image parts on every adapter.

**Changed**
- `src/chat/tool/tool_response.rs`: `parts` field + builders.
- Native image serialization where the wire supports it: Anthropic `tool_result` image blocks, Bedrock Converse `toolResult` image blocks, OpenAI Responses `function_call_output` output arrays.
- Follow-up user-message carriage elsewhere: OpenAI Chat Completions-compatible providers (batched across a tool-message run), Gemini (after the merged `functionResponse` turn), Ollama (native `images` array); `"(see attached image)"` placeholder for image-only results.
- Shared helpers in `adapters/support.rs` (`TOOL_RESULT_IMAGES_LABEL`, `tool_response_fallback_text`). Non-image parts are skipped with a warning; text-only responses keep byte-identical previous output. CHANGELOG + `dev/specs/spec-tool.md` / `spec-chat.md`.

**Breaking**: downstream `ToolResponse` struct literals must add `parts: None` (or use the builders).

**Foundation for the rest of the split**: the next two PRs (embedded tool responses, Anthropic URL images) build on `ToolResponse.parts` and these helpers, and will follow after this.

**Tests**: build, `test --lib` (219), `test --lib --features otel` (241), `fmt`, `clippy` clean; the five yakbak replay suites pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,13 @@
 
 - `!` API CHANGE - `Error::HttpError` adds a `headers: Box<HeaderMap>` field carrying the response headers of failed streaming HTTP calls.
 - `!` API CHANGE - `Tool` adds the public `custom_format: Option<Value>` field for provider-native freeform custom-tool formats. Downstream `Tool` struct literals must add `custom_format: None`, or preferably migrate to `Tool::new(...)` and builder methods. `Tool::with_custom_format(...)` is the new builder API.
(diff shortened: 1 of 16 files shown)


jeremychone:
@VikashLoomba  Good PR, but, since we are doing a version update, 
we can probably refactor the tool content like this: 

```
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct ToolResponse {
    pub call_id: String,
    pub fn_name: Option<String>,
    pub content: ToolResponseContent,
}

#[derive(Debug, Clone, Default)]
pub struct ToolResponseContent {
    parts: Vec<ToolResponsePart>,
}

#[derive(Debug, Clone, Serialize, Deserialize)]
pub enum ToolResponsePart {
    Text(String),
    Binary(Binary),
}
```

And we use similar builders and extractors we have for message content. 

What do you think? 

@ropoctl what do you think?

PR #282, opened 2026-08-15
feat(anthropic): preserve signed thinking blocks for replay

chaizhenhua:
## Summary

- add a replayable ThinkingBlock that keeps Anthropic thinking text bound to its optional signature
- parse signed thinking in non-streaming responses and accumulate split thinking/signature streaming deltas into the same block
- preserve the original Thinking, Text, and ToolCall block order in captured MessageContent
- serialize assistant Thinking blocks back onto the Anthropic Messages request wire
- consolidate stream terminal capture around one ordered MessageContent source of truth

## Test design

The regression cases cover both response modes and the replay boundary:

- non-streaming Thinking -> Text -> ToolCall parsing and assistant request round-trip
- streaming split thinking/signature deltas -> ordered terminal Thinking -> Text -> ToolCall capture
- legacy normalized reasoning_content remains available without replacing ordered content

## Validation

- cargo check --all-targets
- cargo test --lib: 200 passed
- all offline Yakbak suites: 25 passed, 10 ignored
- cargo clippy --all-targets -- -D warnings
- cargo fmt --all -- --check

A downstream DeepSeek Anthropic Messages integration also completed a real signed-thinking tool continuation using the captured assistant turn.

### DIFF
--- a/tests/data/yakbak/anthropic/thinking_tool_stream/response_000.txt
+++ b/tests/data/yakbak/anthropic/thinking_tool_stream/response_000.txt
@@ -1,0 +1,44 @@
+event: message_start
+data: {"type":"message_start","message":{"id":"msg_thinking_tool","type":"message","role":"assistant","content":[],"model":"deepseek-v4-pro","stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":31,"output_tokens":2,"cache_creation_input_tokens":0,"cache_read_input_tokens":0}}}
+
+event: content_block_start
+data: {"type":"content_block_start","index":0,"content_block":{"type":"thinking","thinking":"","signature":""}}
+
+event: content_block_delta
+data: {"type":"content_block_delta","index":0,"delta":{"type":"thinking_delta","thinking":"I should "}}
+
+event: content_block_delta
+data: {"type":"content_block_delta","index":0,"delta":{"type":"thinking_delta","thinking":"use the fixture."}}
+
+event: content_block_delta
+data: {"type":"content_block_delta","index":0,"delta":{"type":"signature_delta","signature":"sig-"}}
+
+event: content_block_delta
+data: {"type":"content_block_delta","index":0,"delta":{"type":"signature_delta","signature":"123"}}
+
+event: content_block_stop
+data: {"type":"content_block_stop","index":0}
+
+event: content_block_start
+data: {"type":"content_block_start","index":1,"content_block":{"type":"text","text":""}}
+
+event: content_block_delta
+data: {"type":"content_block_delta","index":1,"delta":{"type":"text_delta","text":"Checking."}}
+
+event: content_block_stop
+data: {"type":"content_block_stop","index":1}
+
+event: content_block_start
+data: {"type":"content_block_start","index":2,"content_block":{"type":"tool_use","id":"toolu_stream_1","name":"read_fixture","input":{}}}
+
+event: content_block_delta
+data: {"type":"content_block_delta","index":2,"delta":{"type":"input_json_delta","partial_json":"{\"marker\":\"anthropic\"}"}}
+
+event: content_block_stop
+data: {"type":"content_block_stop","index":2}
+
+event: message_delta
+data: {"type":"message_delta","delta":{"stop_reason":"tool_use","stop_sequence":null},"usage":{"output_tokens":27}}
+
+event: message_stop
+data: {"type":"message_stop"}
--- a/src/adapter/adapters/anthropic/adapter_shared.rs
+++ b/src/adapter/adapters/anthropic/adapter_shared.rs
@@ -7,7 +7,7 @@ use crate::adapter::{Adapter, AdapterKind, ServiceType, WebRequestData};
 use crate::chat::{
 	Binary, BinarySource, CacheControl, CacheCreationDetails, ChatOptionsSet, ChatRequest, ChatResponse,
 	ChatResponseFormat, ChatRole, ContentPart, JsonSchemaDialect, MessageContent, PromptTokensDetails, ReasoningEffort,
-	StopReason, Tool, ToolCall, ToolChoice, ToolConfig, ToolName, Usage, sanitize_json_schema,
+	StopReason, ThinkingBlock, Tool, ToolCall, ToolChoice, ToolConfig, ToolName, Usage, sanitize_json_schema,
 };
 use crate::resolver::{AuthData, Endpoint};
 use crate::webc::{WebClient, WebResponse};
@@ -235,6 +235,7 @@ impl AnthropicAdapter {
 								}
 								ContentPart::ThoughtSignature(_) => {}
 								ContentPart::ReasoningContent(_) => {}
+								ContentPart::Thinking(_) => {}
 								ContentPart::Custom(custom_part) => values.push(custom_part.data),
 							}
 						}
@@ -272,6 +273,16 @@ impl AnthropicAdapter {
 									"name": tool_call.fn_name,
 									"input": input,
 								}));
+							}
+							ContentPart::Thinking(block) => {
+								let mut value = json!({
+									"type": "thinking",
+									"thinking": block.thinking,
+								});
+								if let Some(signature) = block.signature {
+									value.x_insert("signature", signature)?;
+								}
+								values.push(value);
 							}
 							// Unsupported for assistant role in Anthropic message content
 							ContentPart::Binary(_) => {}
@@ -311,6 +322,7 @@ impl AnthropicAdapter {
 								}));
 							}
 							ContentPart::Custom(custom_part) => values.push(custom_part.data),
+							ContentPart::Thinking(_) => {}
 							_ => {}
 						}
 					}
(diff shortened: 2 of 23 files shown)


jeremychone:
@chaizhenhua is that the same thing as #275  ?

tests/data/yakbak/anthropic/thinking_tool_stream/response_000.txt
```diff
@@ -0,0 +1,44 @@
+event: message_start
+data: {"type":"message_start","message":{"id":"msg_thinking_tool","type":"message","role":"assistant","content":[],"model":"deepseek-v4-pro","stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":31,"output_tokens":2,"cache_creation_input_tokens":0,"cache_read_input_tokens":0}}}
```
ropoctl, line 2: is this an "extension" of anthropic api that deepseek has, or is it documented by anthropic

ropoctl:
Two behavior changes not mentioned in the PR body

- Text capture is no longer concatenated. Previously the Anthropic streamer joined every text delta into one string → one Text part. Now each content_block becomes its own ContentPart::Text. Any consumer calling captured_content.first_text() on a response with interleaved text/thinking/text blocks now gets only the first block.
- into_assistant_message_for_tool_use changed for every provider. It used to rebuild a message from thought-signatures + tool calls only; now it passes the whole captured content through, so assistant text is included in tool-use replays across OpenAI, Gemini, Bedrock, etc. Probably an improvement, but it's a cross-provider semantic change buried in a PR titled feat(anthropic).

jeremychone:
@ropoctl , thanks so much for this review.

Ok, those two points are important changes, and I am not sure we want them. 

- `Text capture is no longer concatenated. `, the goal of ContentPart was not to be a part by stream token(s). So this is probably not what we want. The ContentPart construct is to reflect similar constructs sent by the provider, not a way to store tokens. 


- into_assistant_message_for_tool_use, I am not sure about this one. But good context. 


So, @ropoctl , how is this related to your #275?

ropoctl:
> @chaizhenhua is that the same thing as #275 ?

same problem different solution. mine added a ThinkingBlock but didnt add it to the public api as a content type. this one promotes it to ContentPart.

verbose ai slop follows
```1. Is 282 the same as 275?

Same problem, different solution — and it's not derived from yours. Yours opened 2026-08-01; 282 opened 2026-08-15 by a different author. Independent competing fix.

┌──────────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────┐
│                  │                                                    275 (yours)                                                    │                                         282                                         │
├──────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤
│ Size             │ 27 files, +2179/−49                                                                                               │ 23 files, +416/−140                                                                 │
├──────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤
│ Representation   │ Reuses ThoughtSignature + ReasoningContent, paired positionally, AnthropicThinkingBlockMismatch on count mismatch │ New ContentPart::Thinking(ThinkingBlock) variant                                    │
├──────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤
│ Streaming        │ Normalizes start-only / split / duplicate / cumulative signature-delta forms                                      │ Accumulates start + deltas, no normalization of the cumulative case                 │
├──────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────┤
│ Terminal capture │ Keeps InterStreamEnd's split accumulators                                                                         │ Replaces them with one ordered MessageContent + an assemble_captured_content helper │
└──────────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────┘

One concrete collision: 282 adds tests/data/yakbak/anthropic/thinking_tool_stream/response_000.txt as a new file. That path doesn't exist on origin/main — yours creates it too. Whichever merges first leaves the other with an add/add conflict on a cassette, which git can't auto-resolve.

2. Official Anthropic API or DeepSeek extension?

The block shape is official; the optional signature is the DeepSeek accommodation.

{"type": "thinking", "thinking": "...", "signature": "..."} is the real Anthropic Messages API extended-thinking block, and redacted_thinking (with a data field) is its sibling — neither PR handles that one.

What's non-Anthropic is item.x_take::<String>("signature").ok() → Option<String>. Your own recorded cassettes settle this — in every real Anthropic recording in the repo the key is present:

- thinking_tool_non_stream/response_000.txt — full signature (yAIMG875Fo67JvSRNNM4...)
- thinking_tool_stream/response_000.txt, thinking_adjudication_tool_stream/response_001.txt — content_block_start carries {"type":"thinking","thinking":"","signature":""}, then signature_delta fills it

Empty string at block start, never absent. Option only earns its keep for a provider that omits the key. (Your thinking_signature_variants_stream cassette is synthetic, so it isn't evidence either way.)

On the replay side this matters: Anthropic's ThinkingBlockParam requires both thinking and signature, and the signature must be echoed back byte-identical — the API rejects tampering. An unsigned thinking block replayed into an assistant turn is a 400. So Option<String> lets you construct a request Anthropic will reject, where your version errors early.

The provenance tells line up: the PR's test uses model "deepseek-v4-pro" under AdapterKind::Anthropic, and the PR body says "a downstream DeepSeek Anthropic Messages integration." Signatures are Anthropic's own integrity mechanism, so an Anthropic-compatible third-party endpoint has nothing to sign with. That's inference from the PR's own artifacts rather than DeepSeek docs I verified, but it's the only reading that explains the Option.

3. What is ThinkingBlock?

There are now two, and the difference between them is the whole design argument.

282's — public API surface, new file src/chat/content_part/thinking_block.rs:

#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
pub struct ThinkingBlock {
      pub thinking: String,
      #[serde(default, skip_serializing_if = "Option::is_none")]
      pub signature: Option<String>,
}

It becomes ContentPart::Thinking(ThinkingBlock) — a variant in the universal content enum, so every adapter must match on it and every downstream consumer sees it.

Yours — already in your branch at src/adapter/adapters/anthropic/streamer.rs:34, private to the Anthropic streamer, with new(), append_signature_delta(), into_thought_block(). It's a streaming accumulator that normalizes the fragmented/cumulative signature forms, then emits the existing ThoughtSignature + ReasoningContent parts.

So you already built a ThinkingBlock — you just didn't promote it into ContentPart. That's the real fork: 282 makes the bound pair a public content type; you keep the binding as a streamer-internal detail and validate the pairing at serialization.```

**Comment by jeremychone** [comment]
@ropoctl So, what do you think? 

It might be good to expose it so that users can do what they want with it. 

I am not sure, since I do not use those tools too much. So, we have to think it through to have a good, future-proof scheme. 

I would like to do the v0.7.0 soon, but it would be nice to have this type of functionality in since it is an API change. 

Now, I commented on #275 about the opt-in.

**Comment by ropoctl** [comment]
ThinkingBlock as a ContentPart has some value- for anthropic- it 1:1 maps to a thing in that api shape. But how it maps to other ones, less clear. if this library has the goal of making a universal api shape without privileging any particular provider too much, this might not be right. i would say ThinkingBlock, ReasoningContent, ReasoningSignature as 3 separate things in our ContentPart seems ugly. if we add ThinkingBlock, we should subsume the others into it.

edit- no, it would need optionality added to subsume the others:
```
That's also the answer to jeremychone's question 2 — "can this fit across providers?" As written, no. Not because reasoning can't be unified, but because that particular struct encodes Anthropic's binding as mandatory.

The shape that does fit

One variant, both sides optional:

ContentPart::Reasoning(ReasoningPart)

pub struct ReasoningPart {
      /// Human-readable reasoning, when the provider returns it.
      pub text: Option<String>,
      /// Opaque provider token required to replay this turn. Never inspect it.
      pub signature: Option<String>,
      /// Which model/provider issued the signature — signatures are model/provider-bound.
      pub model_iden: Option<ModelIden>,
}

Every current case is expressible without lying about the wire:

- Anthropic → {text: Some, signature: Some} — the binding becomes explicit
- Gemini signature part → {text: None, signature: Some}
- Gemini thought text / DeepSeek reasoning_content → {text: Some, signature: None}
- openai_resp → {text: None, signature: Some(encrypted_content)}, matching what it already emits
```

jeremychone:
Ok, so perhaps we keep that internal for now, like #275, and then see if things become more obvious over time.

@chaizhenhua feel free to chime in.

chaizhenhua:
Thanks — after comparing the two, I agree they’re solving the same underlying problem with different API designs. Since the current preference is to keep the Anthropic-specific thinking/signature pairing internal for now, #275 is the better fit.

I’ll close this in favor of #275 and check whether it is missing any useful coverage from this PR.

jeremychone:
@chaizhenhua Thanks for Chiming in. We will work on #275

jeremychone:
@chaizhenhua so PR #275 was merged. Let us know if it meets your need.

jeremychone:
@chaizhenhua I am going to close this one since it should have been fixed in #275 

feel free to open issue/pr if something is missing.

PR #284, opened 2026-08-17
Built-in tool fixes for Gemini

vagmi:
When doing some token accounting on my end I found that the Gemini adapter ignores tokens that are obtained through in-built tools like webSearch and urlContext. Only server-side tools are populated under the `toolUsePromptTokenCount` [1]. I also found another case where we need to set the `include_server_side_tool_invocations` [2] to enable mixing built-in tools like urlContext and user specified tools. I have recorded a yakbak casettes for both the cases. 

[1]: https://docs.cloud.google.com/gemini-enterprise-agent-platform/reference/rest/v1/GenerateContentResponse#UsageMetadata
[2]: https://ai.google.dev/gemini-api/docs/generate-content/tool-combination#enable


jeremychone:
pr commit message
```
* Record tool use token count for gemini along with input tokens

* Allow mixing builtin and user supplied tools.
```

jeremychone:
@vagmi Thanks, good fix! Clean PR.

PR #285, opened 2026-08-18
fix(chat): default ChatOptions.stop_sequences for partial deserialization

ropoctl:
## Problem

Every field of `ChatOptions` is optional in practice, and serde treats the `Option<T>` ones as such automatically. `stop_sequences` is the one exception — a bare `Vec<String>` — so deserializing anything that omits it fails:

```rust
let options: ChatOptions = serde_json::from_value(json!({"temperature": 0.4}))?;
// Error("missing field `stop_sequences`")
```

This makes a partial `ChatOptions` impossible to deserialize from a config file, or from a language binding that accepts an options mapping — even though `ChatOptions::default()` fills the field with an empty `Vec` perfectly happily.

I hit this in Python bindings that accept either a typed options object or a plain dict: `{"temperature": 0.4}` had to be padded with `"stop_sequences": []` to deserialize at all.

## Change

Adds `#[serde(default)]` to the field, matching what `ChatRequest.messages` already does for the same reason, plus two regression tests.

## Verification

- Both new tests fail without the one-line change (`Error("missing field 'stop_sequences'")`) and pass with it.
- Full `cargo test --lib` suite passes (201 tests).
- `cargo fmt --check` clean.

No behaviour change for any input that already deserialized, and none for serialization.

### DIFF
--- a/src/chat/chat_options.rs
+++ b/src/chat/chat_options.rs
@@ -26,6 +26,7 @@ pub struct ChatOptions {
 	pub top_p: Option<f64>,
 
 	/// Sequences that halt generation when encountered.
+	#[serde(default)]
 	pub stop_sequences: Vec<String>,
 
 	// -- Stream Options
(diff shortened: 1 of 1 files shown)


jeremychone:
PR commit message
```
Every field of ChatOptions is optional in practice, and serde treats the
Option<T> ones as such automatically — but `stop_sequences` is a bare
Vec<String>, so deserializing anything that omits it fails:

    let options: ChatOptions = serde_json::from_value(json!({"temperature": 0.4}))?;
    // Error("missing field `stop_sequences`")

That makes it impossible to deserialize a partial ChatOptions from config
or from a client binding that accepts an options mapping, even though the
struct's own Default fills the field with an empty Vec.

Adds #[serde(default)], matching what ChatRequest.messages already does
for the same reason, plus two regression tests.
```

jeremychone:
@ropoctl Thanks for the focused fix. Good one.

Now, a related thing with `ChatOptions`, one thing I really do not like is that it serializes `None` and the empty `Vec` as `null` and `[]`. This adds a lot of unnecessary noise.

So, I will do a separate PR to add `skip_serialize` for `None` or an empty `Vec` in `ChatOptions`.

That's my default, but I accepted this PR a long time ago and did not add that.

PR #286, opened 2026-08-18
adapter: AdapterKind::all() to enumerate the built-in adapters

ropoctl:
## Problem

`AdapterKind` can be parsed from a string (`from_lower_str`) and rendered back to one (`as_lower_str`), but there is no way to ask *which adapters exist*.

Anything that needs the set has to hand-maintain a duplicate list:

- validating a configuration value against the known providers, with an error message that names the valid options
- offering a provider choice in a UI or CLI
- building a lookup keyed by adapter kind

I hit this writing Python bindings: to give `AdapterKind("not-a-provider")` an error message listing the valid names, I ended up with a 29-entry table copied out of `adapter_kind_str_maps!` plus a round-trip test whose only job is to fail when this crate renames something. That table should not exist.

## Change

The `adapter_kind_str_maps!` table already names every variant, so `all()` falls out of the existing macro — no second source of truth:

```rust
pub fn all() -> &'static [AdapterKind]
```

`Custom(_)` is excluded, since it is parameterized by an index rather than a fixed variant and cannot be enumerated. `BedrockSigv4` is included under its existing `#[cfg]`, matching how the other generated methods handle it.

## Verification

Three tests: every entry round-trips through `from_lower_str`, the list has no duplicates, and `Custom` is absent.

Full `cargo test --lib` passes (202 tests), `cargo fmt --check` clean. Purely additive.

### DIFF
--- a/src/adapter/adapter_kind.rs
+++ b/src/adapter/adapter_kind.rs
@@ -273,3 +273,47 @@ impl AdapterKind {
 		}
 	}
 }
+
+// region:    --- Tests
+
+#[cfg(test)]
+mod tests {
+	type Result<T> = core::result::Result<T, Box<dyn std::error::Error>>; // For tests.
+
+	use super::*;
+
+	#[test]
+	fn test_adapter_kind_all_round_trips_lower_str() -> Result<()> {
+		// -- Exec & Check
+		for kind in AdapterKind::all() {
+			let lower = kind.as_lower_str();
+			assert_eq!(AdapterKind::from_lower_str(lower), Some(*kind), "for '{lower}'");
+		}
+
+		Ok(())
+	}
+
+	#[test]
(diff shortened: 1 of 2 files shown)


jeremychone:
@ropoctl Thanks, clean, simple, and nice addition. Thanks!

jeremychone:
PR commit message

```
AdapterKind can be parsed from a string and rendered back to one, but
there is no way to ask which adapters exist. Anything that needs the set —
validating a config value against the known providers, offering a choice,
or building a lookup keyed by adapter — has to hand-maintain a duplicate
list and keep it in sync by hand.

The adapter_kind_str_maps! table already names every variant, so the list
comes out of the existing macro for free.

Custom(_) is excluded: it is parameterized by an index rather than a fixed
variant, so it cannot be enumerated. Tests cover the round-trip through
from_lower_str, uniqueness, and that exclusion.
```

PR #287, opened 2026-08-18
error: Error::status() accessor

ropoctl:
## Problem

The HTTP status a provider responded with is already carried by the error, but reaching it requires knowing the internal shape:

```rust
match &err {
    Error::WebModelCall { webc_error: webc::Error::ResponseFailedStatus { status, .. }, .. } => Some(*status),
    Error::WebAdapterCall { webc_error: webc::Error::ResponseFailedStatus { status, .. }, .. } => Some(*status),
    Error::HttpError { status, .. } => Some(*status),
    _ => None,
}
```

Three variants, two layers, and the `WebAdapterCall` arm is easy to miss entirely — it only shows up on `all_model_names`, so a caller who tests against chat calls will ship without it.

Every consumer that wants to distinguish "rate limited, back off" from "bad request, fix it" has to write this, and any that gets it subtly wrong fails in exactly the situation it was written for.

## Change

```rust
impl Error {
    pub fn status(&self) -> Option<StatusCode>
}

impl webc::Error {
    pub fn status(&self) -> Option<StatusCode>
}
```

`None` for failures with no HTTP response behind them: connection and timeout errors, resolver failures, request validation, stream parsing.

Deliberately *not* included: any notion of which statuses are retryable. That is policy and belongs to the application — this only reports what the provider said. It pairs with the response headers `HttpError` already carries (#280), so a retry layer can read both the status and `retry-after` without matching on variants.

## Verification

Five tests covering `WebModelCall`, `WebAdapterCall`, `HttpError`, a non-HTTP error, and a non-status `webc::Error`.

Full `cargo test --lib` passes, `cargo fmt --check` clean. Purely additive.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -170,6 +170,33 @@ Drop the `::` namespace prefix or `ModelSpec::Iden`, or build a Client without \
 	SerdeJson(serde_json::Error),
 }
 
+/// Accessors
+impl Error {
+	/// The HTTP status the provider responded with, when the failure came
+	/// from an HTTP response.
+	///
+	/// The status is already carried by the error, but reaching it means
+	/// knowing that a failed chat call arrives as
+	/// `WebModelCall { webc_error: webc::Error::ResponseFailedStatus { .. } }`,
+	/// that adapter-level calls use `WebAdapterCall` instead, and that
+	/// `HttpError` is a third, separate shape. Callers that want to branch
+	/// on 429 or 5xx should not have to know any of that.
+	///
+	/// Returns `None` for failures with no HTTP response behind them:
+	/// connection and timeout errors, resolver failures, request
+	/// validation, stream parsing.
+	///
+	/// Retry policy stays with the caller — this only reports what the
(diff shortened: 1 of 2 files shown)


jeremychone:
PR commit message

```
The HTTP status a provider responded with is already carried by the error,
but reaching it means knowing that a failed chat call arrives as
WebModelCall { webc_error: ResponseFailedStatus { .. } }, that adapter-level
calls use WebAdapterCall instead, and that HttpError is a third, separate
shape. A caller that just wants to branch on 429 or 5xx has to match all
three and reach through the webc layer to do it.

Adds Error::status() -> Option<StatusCode>, and webc::Error::status()
underneath it, returning None for failures with no HTTP response behind
them: connection and timeout errors, resolver failures, request validation,
stream parsing.

This is fact extraction only — no retry or backoff policy, which stays with
the application. It pairs with the response headers HttpError already
carries, so a retry layer can read both the status and retry-after without
matching on error variants.
```

PR #288, opened 2026-08-18
fix(client): apply the ServiceTargetResolver when resolving adapter config

ropoctl:
## Problem

`Client::all_model_names()` resolves its endpoint and auth through `ClientConfig::resolve_adapter_config`, which runs the **AuthResolver** but not the **ServiceTargetResolver**:

```rust
pub(crate) async fn resolve_adapter_config(&self, adapter_kind: AdapterKind) -> Result<(AuthData, Endpoint)> {
    let model = ModelIden::new(adapter_kind, "");
    let auth = self.run_auth_resolver(model).await?;
    let endpoint = AdapterDispatcher::default_endpoint(adapter_kind);   // <-- resolver never runs
    Ok((auth, endpoint))
}
```

A Client configured to reach a private endpoint that way — a local vLLM, an OpenAI-compatible gateway — therefore lists models from the adapter's **public default** instead, and sends it the credential the AuthResolver just supplied for the private host.

It fails in the worst direction: silently, and with a successful-looking result. I found it because a test suite that should have been fully offline made a live request to `api.openai.com`; the client under test had been configured with a `ServiceTargetResolver` pointing at a local fake, and `all_model_names` ignored it.

```rust
    .with_auth_resolver(/* key for the private gateway */)
    .with_service_target_resolver(/* endpoint: https://gateway.internal/v1/ */)
    .build();

client.all_model_names(AdapterKind::OpenAI, ()).await?;
// -> queries https://api.openai.com/v1/models, with the gateway's key
```

## Change

Build the `ServiceTarget` the same way `resolve_service_target` does and run the resolver, then take its auth and endpoint.

Both resolvers already see a `ModelIden` with an empty model name here — listing models is not about any one model, and that convention is already established for the AuthResolver on the line above. This just extends the same treatment to the second resolver.

## Verification

Two tests, using the existing `bound_config` helper: a configured resolver is honored, and the adapter default still stands when no resolver is attached. The first fails on `main` (`left: "https://api.openai.com/v1/"`, `right: "https://custom.example/v1"`) and passes with the change.

Full `cargo test --lib` passes (201 tests), `cargo fmt --check` clean.

## Note on scope

The alternative reading is that `all_model_names` already takes a `ProviderConfig` argument, so callers *can* pass the endpoint explicitly. But the default path silently disagreeing with how every other call on the same Client routes is the surprising part — and the failure mode is credentials going to an unintended host, which seems worth closing regardless.

### DIFF
--- a/src/client/config.rs
+++ b/src/client/config.rs
@@ -132,11 +132,21 @@ impl ClientConfig {
 	/// Resolves auth and endpoint for the given adapter kind.
 	///
 	/// Used by `Client::all_model_names()` where no specific model name is available.
+	///
+	/// Both resolvers see a `ModelIden` with an empty model name, since
+	/// listing models is not about any one model. The `ServiceTargetResolver`
+	/// is run for the same reason the `AuthResolver` is: a Client pointed at
+	/// a private endpoint must not fall back to the adapter's public default
+	/// here, or it would send that Client's credentials to the wrong host.
 	pub(crate) async fn resolve_adapter_config(&self, adapter_kind: AdapterKind) -> Result<(AuthData, Endpoint)> {
 		let model = ModelIden::new(adapter_kind, "");
-		let auth = self.run_auth_resolver(model).await?;
+		let auth = self.run_auth_resolver(model.clone()).await?;
 		let endpoint = AdapterDispatcher::default_endpoint(adapter_kind);
-		Ok((auth, endpoint))
+
+		let service_target = ServiceTarget { model, auth, endpoint };
+		let service_target = self.run_service_target_resolver(service_target).await?;
+
+		Ok((service_target.auth, service_target.endpoint))
 	}
 
 	/// Resolves a ServiceTarget for the given model.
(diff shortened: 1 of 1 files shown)


jeremychone:
@ropoctl Ok, I will have to look at this. The problem, as you mentioned, is that resolving the server target gets the model ID, which is not needed when listing all. This is why I had this Provider concept.

But I understand the concern.

I need to revisit my previous assessment.

ropoctl:
For sequencing: #289 builds on this one (both rewrite `resolve_adapter_config`), so its branch carries this commit as well. Merging this first makes that PR a single-commit diff.

ropoctl:
Correction to my earlier comment: #289 is now a standalone single-commit PR on `main`, not carrying this one. The two do touch the same function, so whichever lands second needs a small rebase — happy to do that in either order.

jeremychone:
PR commit message

```
`Client::all_model_names()` resolves through `resolve_adapter_config`,
which runs the AuthResolver but takes the adapter's default endpoint and
never runs the ServiceTargetResolver.

So a Client configured to reach a private endpoint that way — a local
vLLM, an OpenAI-compatible gateway — lists models from the adapter's
public default instead, and sends the credential the AuthResolver just
supplied to that host. It fails silently in the worst direction: the call
succeeds against the wrong provider.

Both resolvers already see a `ModelIden` with an empty model name here,
since listing models is not about any one model; this just extends the
same treatment to the second resolver, and builds the ServiceTarget the
same way `resolve_service_target` does.

Tests cover both directions: a configured resolver is honored, and the
adapter default still stands when no resolver is attached.
```

jeremychone:
@ropoctl Merged, thanks.

PR #289, opened 2026-08-18
client: ClientBuilder::with_provider_config for per-adapter endpoint/auth

ropoctl:
> **Ordering note:** this touches `resolve_adapter_config`, as does #288, so
> whichever lands second needs a trivial rebase — the two changes are adjacent,
> not contradictory (#288 makes that function run the `ServiceTargetResolver`;
> this one makes the endpoint it starts from configurable). Happy to rebase on
> request, in either order. Kept as its own PR rather than bundling the two,
> since they are independent changes; GitHub's stacked PRs can't help here, as
> they ["require all branches to be in the same repository. Cross-fork stacks
> are not supported."](https://docs.github.com/en/pull-requests/get-started/about-stacked-prs)

## Problem

`ProviderConfig` already exists, and `all_model_names` already takes one — but there is no way to attach one to a `Client`.

So pointing an adapter at a different endpoint with a different key (a gateway, a local vLLM, a regional deployment) means hand-writing two resolvers whose entire job is to compare adapter kinds:

```rust
let key = key.clone();
builder = builder
    .with_auth_resolver(AuthResolver::from_resolver_fn(move |iden: ModelIden| {
        if iden.adapter_kind == kind { Ok(Some(AuthData::from_single(key.clone()))) }
        else { Ok(None) }
    }))
    .with_service_target_resolver(ServiceTargetResolver::from_resolver_fn(
        move |mut t: ServiceTarget| {
            if t.model.adapter_kind == kind { t.endpoint = endpoint.clone(); }
            Ok(t)
        },
    ));
```

And because a `Client` holds one resolver of each kind, doing it for a *second* provider means folding both into a single closure that dispatches on adapter kind — reimplementing, in every downstream project, the lookup table this change stores directly.

## Change

```rust
Client::builder()
    .with_provider_config(
        AdapterKind::OpenAI,
        (Endpoint::from_static("https://gateway.internal/v1/"), AuthData::from_env("GATEWAY_KEY")),
    )
    .with_provider_config(AdapterKind::Ollama, Endpoint::from_static("http://localhost:11434/v1/"))
    .build()
```

Repeatable, and it accepts anything `Into<ProviderConfig>` — so the existing `Endpoint`, `AuthData`, and tuple conversions all work.

**Precedence, lowest to highest: the adapter's built-in default → this configuration → the resolvers.** Resolvers keep the final say, so nothing that works today changes behaviour; configuring one provider never affects another.

It seeds both `resolve_service_target` and `resolve_adapter_config`, so chat calls and `all_model_names` agree on where a given provider lives.

## Verification

Four tests: endpoint and auth are supplied; configuration is scoped to its own adapter; a resolver still overrides it; and it applies to the no-model `resolve_adapter_config` path too.

`cargo test --lib` passes (203 tests) and `cargo fmt --check` is clean. The new doctest passes; the two failing doctests (`Tool::schema`, `zai`) already fail on `main` and are untouched here.

### DIFF
--- a/src/client/builder.rs
+++ b/src/client/builder.rs
@@ -1,7 +1,7 @@
 use crate::adapter::AdapterKind;
 use crate::chat::ChatOptions;
 use crate::resolver::{
-	AuthResolver, IntoAuthResolverFn, IntoModelMapperFn, IntoServiceTargetResolverFn, ModelMapper,
+	AuthResolver, IntoAuthResolverFn, IntoModelMapperFn, IntoServiceTargetResolverFn, ModelMapper, ProviderConfig,
 	ServiceTargetResolver,
 };
 use crate::webc::WebClient;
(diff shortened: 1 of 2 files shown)


jeremychone:
@ropoctl interesting, I will need to look further. But interesting one.

jeremychone:
@ropoctl I like this approach. It fits well into what we have. Nice find. 

One very minor point is that since ProviderConfig is a list, we should not use `with_provider_config(...)`, but `append_provider_config(...)`. 

This way it follows my/rust10x [fluid apis](https://github.com/rust10x/rust10x/blob/main/pack/guide/base/pattern-fluid-and-builder.md). 

Does that make sense? Also, there is some conflict issue with the previous merge.

Once `append_provider_config` and the conflict are resolved, I will merge it.

jeremychone:
@ropoctl Ok, seems you did the change. Let me know if all ok for me to merge.

ropoctl:
should be good to go

jeremychone:
PR commit message: 

```
ProviderConfig already exists and all_model_names already takes one, but
there is no way to attach one to a Client. Pointing an adapter at a
different endpoint with a different key — a gateway, a local vLLM, a
regional deployment — means hand-writing an AuthResolver and a
ServiceTargetResolver whose only job is to compare adapter kinds.

Since a Client holds one resolver of each kind, doing this for a second
provider means folding both into a single closure that dispatches on
adapter kind — which is the same table this change stores directly.

    Client::builder()
        .append_provider_config(AdapterKind::OpenAI, (endpoint, AuthData::from_env("GATEWAY_KEY")))
        .append_provider_config(AdapterKind::Ollama, Endpoint::from_static("http://localhost:11434/v1/"))
        .build()

Precedence, lowest to highest: adapter default, then this configuration,
then the resolvers — so resolvers keep the final say and configuring one
provider never affects another. It seeds resolve_service_target and
resolve_adapter_config alike, so chat and model listing agree on where a
provider lives.
```

jeremychone:
released in `0.7.0-beta.21`

PR #290, opened 2026-08-18
Add mechanism to observe chat stream frames across providers

vagmi:
Right now, the `exec_chat` call allows you to capture the raw response but `exec_chat_stream` does not afford that. Sometimes you need access to frames for counting search queries or debug SSE frames. The `ChatFrameSink` trait allows us to implement provider specific escape hatches without having to change the core types. I've done my best to ensure that I borrow and not make unnecessary allocations. We still need the `Arc<dyn ChatFrameSink>` to ensure that `ChatOptions` does not get complex with generics and it keeps the API backwards compatible. I feel this is a good compromise for accommodating provider specific accounting without polluting the API. In my case, I wanted to account for web search queries as each grounding search query costs $0.014 on gemini and $0.01 on OpenAI and Anthropic when executed via server tools. This can also be very useful for debugging and tracing stream frames.


jeremychone:
@vagmi Okay, interesting and good feature. 

I did a review with AI (pro@coder agent, so quite 'in the loop' review). Here is the thing that came back (see below)


Does this make sense to you? You thought more about this topic, so I am definitely interested to see your perspective on each point. 

The only thing is that this library should never have `.expect(..)` or `.unwrap()` in the prod code path (i.e. outside of examples and tests).

I think I have some there, and I need to clean them up, but I want to make sure PRs do not introduce any.

JC Assisted AI review answer: 

```

### Key Strengths of the Design

1. Clean separation of concerns:
   - `FrameTap` is embedded in the low-level transports (`EventSourceStream`, `WebStream`, and `BedrockStreamer`), leaving adapters mostly unchanged.
   - `ChatStream` coordinates lifecycle callbacks (`on_end`, `on_error`) using the high-level `StreamEnd` and typed errors.
2. Zero allocation overhead for passive frames:
   - `RawFrameRef<'a>` borrows wire data directly. No JSON parsing or string allocation happens unless the sink explicitly calls `.json()` or `.to_owned_frame()`.
3. Provider coverage:
   - Works across SSE formats (OpenAI, Anthropic, Gemini), line-delimited JSON (Ollama, Cohere), and binary AWS event-stream framing (Bedrock).
4. Sensible provided sinks:
   - `FnSink` for simple inline inspection.
   - `CollectorSink` for test verification and debugging.
   - `ChannelSink` using non-blocking `try_send` with a dropped frame counter to protect the stream poll loop from stalling.

### Agreement and Nuance on Identified Risks

1. Terminal callback guarantees:
   - I agree with the initial assessment that `ChatStream` should enforce a single-fire latch for terminal callbacks. If `poll_next` encounters multiple errors, `on_error` should only fire once.
   - For abrupt stream drops or premature upstream EOF where no `End` or `Err` event is parsed, the sink will not receive `on_end`. This should be explicitly documented as expected behavior for aborted/dropped client streams.
2. Public API surface:
   - `ChatOptions::raw_frame_sink` is marked `#[serde(skip)]`, so JSON serialization compatibility is preserved.
   - Direct struct literal construction without `..Default::default()` could theoretically fail, but `ChatOptions` is consistently used via fluent builder methods (`with_...`) in idiomatic `genai` code.
   - `ChannelSink` exposes Tokio's `mpsc::Sender`. Because `genai` already depends on Tokio as its primary async runtime, this does not add an unwanted dependency.

### Recommended Next Steps Before Merging

a. Add a terminal state latch in `ChatStream` / `FrameTap` to guarantee `on_end` and `on_error` are mutually exclusive and fire at most once.
b. Replace any `.expect(...)` calls in new unit and integration tests with `.ok_or(...)?` or `?` per project guidelines.
c. Document in `ChatFrameSink` that `on_end` fires only upon clean completion of a stream and is skipped if the stream is dropped prematurely.
d. Polish doc comments and minor typographical errors in `src/chat/frame_sink.rs`.
```

vagmi:
@jeremychone - Sure thing. I think having a terminal state of either error or end is a good call. There are no `unwrap` or `except`s in the prod path or in the tests. I fixed the doc comments as well.

jeremychone:
Commit message from PR
```
* Add mechanism to observe chat stream frames across providers

Right now exec_chat allows you to capture the entire body but
exec_chat_stream does not afford that. Sometimes you need access to
frames for counting search queries or debug server frames. The
ChatFrameSink trait allows us to implement provider specific escape
hatches without having to change the core types.

* Make on_error and on_end mutually exclusive. Fix language on some of the doc comments
```

jeremychone:
@vagmi Merged, thank you for this add and good code quality.

PR #293, opened 2026-08-31
Add support for Gemini Interactions API

vagmi:
Google has its own flavor of the responses API and has started adding models exclusively behind the interactions API e.g., `gemini-3.5-transcribe`.  

> The Interactions API is the best way to build with Gemini models and agents. As of June 2026, it is Generally Available and recommended for all new projects. While it is now considered legacy, the original [generateContent](https://ai.google.dev/gemini-api/docs/generate-content/text-generation) API remains fully supported.

https://ai.google.dev/gemini-api/docs/interactions-overview

<del>I have moved all gemini-3.x models to the interactions API just like how we default to responses API for gpt-5x models. Embed and 2.x variations of the models still resolve to the generateContent and embedContent APIs.</del>

You opt in to the interactions API with the `gemini_ix::` prefix.

As a part of this API, I also have added a small example to enrich with provider specific data like google search grounding results. This can be a good way to expose provider specific enrichments without paying to cost for a complicate API surface. See [examples/c99-gemini-search-citations.rs](https://github.com/vagmi/rust-genai/blob/gemini-interactions-api/examples/c99-gemini-search-citations.rs).

AI Disclosure: This code was produced with the help of AI using my coding agent called slop and Claude code but was manually reviewed and edited by me.


jeremychone:
@vagmi This is awesome. 

One little things, since we have `openai_resp`, could we shorten that wone too perhaps `gemini_inter` or something like that 3 to 5 letters max?

jeremychone:
@vagmi Also, perhaps, for the 0.7.0 release, what about keeping the gemini 3.x under the current protocol. 

This way, we minimize the migration risk (we already have a lot in 0.7.0). Then, for gemini 4.x we will default to gemini_inter

vagmi:
Good idea. We can make it an opt in. The codebase already uses `gemini_ix::` as the prefix for choosing the interactions API.

jeremychone:
@vagmi Ha ok, I did not see the `gemini_ix` that's a good name. 

If you can change the package name to reflect that, that would be great. 

And yes, for now, let's be very selective of what gets routed to `gemini_ix`, only the ones we have to for now. 

If users want to use it with other models, they should be able to do `gemini_ix::gemini-3.5-flash-lite` then, it should work. 

Make sense?

vagmi:
@jeremychone - Got it. The adapter package has been renamed to `gemini_ix` as well. The adapter selection is opt in now.  Let me know if this is good.

jeremychone:
@vagmi cool. When you say optin, is with `gemini_nx:....` namespace, right?

vagmi:
@jeremychone - its the `gemini_ix::` namespace.

jeremychone:
Ok, looks good, but one thing I am going to change is that the name of the variant and struct should be `GeminiIx` for symmetry (with `OpenAIResp`

Makes sense? I can make this change after I merge your PR.

vagmi:
@jeremychone - Saved you the trouble. I changed the name of the struct and the variant to `GeminiIx`

jeremychone:
@vagmi Awesome, but do you agree, right? I like the fully explicit names, more Rust-like, but I think, for those, being consistent on the short version is better. Just want to check if I am missing something.

vagmi:
@jeremychone - Yes I agree. I think keeping things consistent is easier for code readability. and `gemini_ix` reads easier too.

jeremychone:
Just to make sure, why do we still have this: `GeminiIx => "GeminiInteractions", "gemini_interactions", adapters::all_adapters::GeminiIxAdapter;`

vagmi:
Oops I missed that one. I took another pass and renamed all the GeminiInteractions to `GeminiIx` for consistency.

jeremychone:
PR Commit comment

```
* Add support for Gemini Interactions API.

https://ai.google.dev/gemini-api/docs/interactions-overview

* Make gemini interactions api an opt in feature.

We still default to the generateContent API and opt in to
interactions API with the gemini_ix prefix.

* Rename adapter from gemini_interactions to gemini_ix

* Changed the variant type to gemini ix

* Rename all Gemini Interactions to GeminiIx
```

jeremychone:
Thanks, merged, and in `0.7.0-beta.21`

Issue #170, opened 2026-03-20
What is the design intent behind the field mismatch in `ContentPartAdded`?


Borber:
https://github.com/jeremychone/rust-genai/blob/f1810c50d49fed5500a489c33c64c717c14e8f90/src/adapter/adapters/openai_resp/streamer.rs#L39-L45

```sh
2026-03-20T04:13:52.347140Z  WARN genai::adapter::adapters::openai_resp::streamer: C:\Users\x\.cargo\git\checkouts\rust-genai-6f8d8b6e4f845c3e\4ebbf2a\src\adapter\adapters\openai_resp\streamer.rs:112: OpenAIRespStreamer - fail to parse event (skipping). Cause: missing field `_output_index`. Data: {"type":"response.content_part.added","content_index":0,"item_id":"msg_01825094915df0390169bcc98062c481908c9068bb603a8c33","output_index":0,"part":{"type":"output_text","annotations":[],"logprobs":[],"text":""},"sequence_number":3}
```

The intention is to define the correct structure; however, since this event has no actual bearing on the processing workflow, is the decision to simply skip parsing it at the deserialization level?

Would it be preferable to use `#[serde(skip_deserializing)]` instead? Additionally, it appears that the current structure does not align with the official API interface.

PR #296, opened 2026-09-07
fix(openai_resp): propagate streaming error events

emecii:
The Responses API can send a top-level `type: "error"` SSE event. The adapter currently treats it as unknown, so callers can receive a successful End after a provider failure. Recognize this event, return the existing `Error::ChatResponse` with the complete JSON payload, and terminate the stream so no success End or subsequent content is emitted.

Adds two loopback replay regressions using the official error protocol example and a partial-output variant, with content capture enabled and disabled. Both fail before the fix. The fixtures are protocol-derived, not live recordings. Related streaming discussion: #170; this does not claim to close that issue's original warning report.

Validation: 306 library tests and all 18 tests across the five CI replay targets pass; default Clippy with warnings denied and changed-file rustfmt pass. Whole-repository formatting differences and an optional `otel` non-exhaustive match error also reproduce on the base revision. No live provider requests were run.

Implemented and tested with assistance from Codex GPT-6.

### DIFF
--- a/src/adapter/adapters/openai_resp/streamer.rs
+++ b/src/adapter/adapters/openai_resp/streamer.rs
@@ -113,6 +113,9 @@ enum RespStreamEvent {
 	#[serde(rename = "response.incomplete")]
 	ResponseIncomplete { response: RespResponse },
 
+	#[serde(rename = "error")]
+	Error,
+
 	#[serde(other)]
 	Unknown,
 }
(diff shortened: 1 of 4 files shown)


jeremychone:
@emecii Thanks! Good fix.

jeremychone:
This PR also has a `other => panic!("expected provider error, got {other:?}"),`

`genai` must be panic free (as much as we can control in this lib)


So, this will require I cleanup later as part of #297. We skip, we ignore, or we make the call fallible, but we cannot panic.

jeremychone:
@emecii Sorry, wrong alert `other => panic!("expected provider error, got {other:?}"),` was on the test_... file, which is kind of what we have throughout the lib. 

So, all good. 

I am changing it in a follow-up PR for the other PR, but the code was good. 

> Note: eventually, I would like to remove all of those `unwrap()`, `expect()` from tests/examples as well, so that it is simpler to check that we do not have any of those, but for now, a lot of test code has those. `ok_or(...)?` is a great way to have clean and simple test/example failures.

PR #297, opened 2026-09-07
fix(bedrock): preserve frame events after start

emecii:
Bedrock event-stream frames can produce more than one internal event. On the
first content or tool frame, the parser emitted `Start` and then returned only
the first item from that frame's event vector, discarding the text delta or
tool-call chunk.

Queue decoded events and drain them over later stream polls. This preserves the
existing event order and emits the terminal event only after queued events have
been delivered.

Adds CPU-only regressions for a text delta and a tool-call start frame. Both
exercise a single raw Bedrock frame and assert `Start`, the frame event, and
`End`.

Validation:

- `cargo test --lib` — 308 passed
- `cargo clippy --lib --tests -- -D warnings`
- `rustfmt --check src/adapter/adapters/bedrock/streamer.rs`
- `git diff --check`

`cargo fmt --check` still reports formatting differences in two unchanged
OpenAI files on the base revision; this PR does not modify them.

Implemented and tested with assistance from Codex GPT-6.

### DIFF
--- a/src/adapter/adapters/bedrock/streamer.rs
+++ b/src/adapter/adapters/bedrock/streamer.rs
@@ -23,6 +23,7 @@ use crate::{Error, ModelIden, Result};
 use bytes::{Buf, BytesMut};
 use futures::Stream;
 use serde_json::{Map, Value};
+use std::collections::VecDeque;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 use value_ext::JsonValueExt;
@@ -36,6 +37,7 @@ pub(super) struct BedrockStreamer {
 	buf: BytesMut,
 	options: StreamerOptions,
 	captured_data: StreamerCapturedData,
+	pending_events: VecDeque<InterStreamEvent>,
 	done: bool,
 	emitted_start: bool,
 	in_progress_tool: Option<ToolCallAccumulator>,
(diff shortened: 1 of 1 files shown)


jeremychone:
@emecii thanks for this fix. Looks good.

Only "minorish" things:

```
1) `src/adapter/adapters/bedrock/streamer.rs` contains two direct `unwrap()` calls:
  - `self.buf[0..4].try_into().unwrap()`
  - `self.buf[4..8].try_into().unwrap()`
Safe because `self.buf.len() < 12`; however, we're trying to avoid any `unwrap()` in production code paths. In fact, I will also update the code to avoid `unwrap`/`expect` in tests as well, but later.

2) `src/adapter/adapters/bedrock/streamer.rs` can potentially panic when calculating and slicing the header and payload ranges. `headers_len` comes from the remote byte stream, and the code does not currently verify that:

  - `headers_len` fits within `total_len`
  - `headers_end <= payload_end`
  - `headers_end <= self.buf.len()`

  A malformed frame could therefore cause a range-index panic at either:

  - `&self.buf[headers_start..headers_end]`
  - `&self.buf[headers_end..payload_end]`

The same parser should validate the frame layout before slicing, including checking that `headers_len <= total_len - 16`. This accounts for the 12-byte prelude and four-byte message CRC.

```

Now, #2 is something that should be fixed eventually.

Anyway, it's robust enough to be merged. Once I finish merging the other one, I might submit a PR to resolve those paths.

So, the approach right now is that either the function is fallible or, if it is not fallible, it cannot panic. We need to find a way to pass and return the empty state.

PR #301, opened 2026-09-07
fix(openai) - capture OpenRouter's reasoning_details and echo them on assistant messages

### LINKED ISSUE #298: bug - OpenAI adapter drops OpenRouter's `reasoning_details` (signed / encrypted reasoning), so a tool round trip through OpenRouter loses the model's continuity token
**Narrow cargo test:** `cargo test --lib openai::adapter_impl::tests::test_to_chat_response_captures_reasoning_details_as_custom_parts` (on the fork branch; fails on main)
**Adapter:** OpenAI / OpenRouter
**Model:** `anthropic/claude-sonnet-4-6` via OpenRouter (any signed- or encrypted-reasoning model)

## What happens

Through OpenRouter, a model whose reasoning carries a continuity token (Anthropic's signed thinking, OpenAI's encrypted reasoning, Gemini's thought signatures) returns it twice: as plaintext in `message.reasoning`, and as a structured entry in `message.reasoning_details`. OpenRouter's docs ([Preserving Reasoning](https://openrouter.ai/docs/use-cases/reasoning-tokens)) say the structured array is what must go back for those models:

> Use `reasoning_details` when working with models that return special reasoning types (such as encrypted or summarized) - this preserves the full structure needed for those models. For models that only return raw reasoning strings, you can use the simpler `reasoning` field. You can also use `reasoning_content` as an alias.

and

> the entire sequence of consecutive reasoning blocks must match the outputs generated by the model during the original request; you cannot rearrange or modify the sequence of these blocks.

The OpenAI adapter reads `/message/reasoning` and `/message/reasoning_content` (text) and echoes `reasoning_content` on assistant messages; nothing in the crate reads or writes `reasoning_details` (`grep -r reasoning_details src/` finds nothing at `0.7.0-beta.21`). So the signed or encrypted entry is dropped on decode and the next request goes out with plaintext only.

## Verified live (2026-09-05)

Turn 1 to `anthropic/claude-sonnet-4-6` via OpenRouter with `reasoning: {max_tokens: 1024}` and a tool returned:

```json
"reasoning": "...(85 chars)...",
"reasoning_details": [{"type": "reasoning.text", "text": "...", "signature": "...", "format": "anthropic-claude-v1", "index": 0}]
```

Replaying the assistant turn plus the tool result three ways:

| Assistant turn replayed as | Result |
|---|---|
| `reasoning_content` text only (what genai sends today) | 200, model answers, **signature not carried** |
| `reasoning_details` verbatim | 200, model answers, signature carried |
| stripped | 200, model answers |

Nothing errors: the gateway accepts the request and the model simply continues without its prior reasoning, which is why this goes unnoticed. `moonshotai/kimi-k3` returns an unsigned `reasoning.text` entry, so for raw-string models the existing `reasoning_content` echo is the documented mechanism and is correct. Sending both `reasoning_content` and `reasoning_details` together on the same assistant message is accepted (200) for both models.

## Proposed change

Carry each `reasoning_details` entry as a `ContentPart::Custom` on decode (verbatim, tagged with the model — the same way an unrecognised Anthropic block is carried), and on the request path echo `Custom` parts whose `type` starts with `reasoning.` back as `reasoning_details`, in order, beside the existing `reasoning_content` echo. Verbatim rather than typed, because the round trip needs `format`, `id` and `index` back unchanged and the sequence untouched. Non-streaming first; `delta.reasoning_details` in the streamer as a follow-up.

A PR implementing this, with unit tests on both paths, is ready: `bricef/rust-genai` branch `feat/openai-reasoning-details`.

bricef:
Fixes #298.

OpenRouter returns a provider's own reasoning blocks — Anthropic's signed text, OpenAI's encrypted content, summaries — as a `reasoning_details` array beside the plaintext `reasoning`, and requires that array back verbatim and in order for a multi-turn tool round trip to keep the model's continuity token; plaintext `reasoning_content` is only enough for models that return raw reasoning strings. The adapter read the plaintext and nothing else, so through OpenRouter a signed or encrypted block was dropped on decode and the next request went out without it. The gateway accepts that request and the model continues without its prior reasoning, so nothing errors.

**Change.** On decode, each `reasoning_details` entry becomes a `ContentPart::Custom` (verbatim, tagged with the model), ahead of the text and tool calls — the same treatment an unrecognised Anthropic block already gets. On encode, `Custom` parts whose `type` starts with `reasoning.` go out again as `reasoning_details`, in order, beside the existing `reasoning_content` echo. Other `Custom` parts stay ignored on this wire. Verbatim rather than typed because the round trip needs `format`, `id` and `index` back unchanged and the sequence untouched.

**Verified live** against `anthropic/claude-sonnet-4-6` and `moonshotai/kimi-k3` through OpenRouter on 2026-09-05: sending both fields together on a replayed assistant turn is accepted, and with `reasoning_details` present the signature is carried; with plaintext only it is not.

**Tests.** `test_to_chat_response_captures_reasoning_details_as_custom_parts` (decode: both entry types, in order, ahead of the tool call, model-tagged, plaintext still in `reasoning_content`), `test_reasoning_details_echoed_verbatim_on_assistant_message` (encode: byte-identical array beside `reasoning_content`), `test_custom_parts_that_are_not_reasoning_blocks_are_not_echoed`. The first two fail without the change.

**Scope.** Non-streaming path only; `delta.reasoning_details` in the streamer is a follow-up. No Anthropic or Gemini behaviour changes. CHANGELOG entry under OpenAI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@
   - `^` Preserve signed thinking blocks and thought signatures across streaming and non-streaming tool turns. (PR #275)
 - OpenAI:
   - `-` Propagate Responses API streaming `error` events as `genai::Error::ChatResponse` and cleanly terminate the stream. (PR #296)
+  - `+` Capture OpenRouter's `reasoning_details` (signed, encrypted and summary reasoning blocks) as `Custom` parts on the non-streaming path and echo them verbatim, in order, on assistant messages, so a multi-turn tool round trip through OpenRouter keeps the model's continuity token; `reasoning_content` is still echoed beside them.
   - `+` Route GPT-6 models (such as `gpt-6-astra`) to the OpenAI Responses API adapter.
   - `+` Support OpenAI Responses freeform custom tools with grammar-constrained raw-string input. Custom tools serialize as `type: "custom"`, custom tool-call input streams incrementally, and round-trips as `custom_tool_call` / `custom_tool_call_output` items. (PR #266)
   - `^` Capture `cache_write_tokens` from prompt-cache usage and normalize it to `Usage.prompt_tokens_details.cache_creation_tokens` for Chat Completions and Responses API payloads.
--- a/src/adapter/adapters/openai/adapter_impl.rs
+++ b/src/adapter/adapters/openai/adapter_impl.rs
@@ -1,7 +1,8 @@
 use super::OpenAIStreamer;
 use crate::adapter::{Adapter, AdapterKind, ServiceType, WebRequestData};
 use crate::chat::{
-	ChatOptionsSet, ChatRequest, ChatResponse, ChatStream, ChatStreamResponse, MessageContent, StopReason, ToolCall,
+	ChatOptionsSet, ChatRequest, ChatResponse, ChatStream, ChatStreamResponse, ContentPart, MessageContent, StopReason,
+	ToolCall,
 };
 use crate::resolver::{AuthData, Endpoint};
 use crate::webc::{EventSourceStream, WebClient, WebResponse};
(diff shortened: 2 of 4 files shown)


jeremychone:
PR commit message:

```
OpenRouter returns a provider's own reasoning blocks — Anthropic's signed
text, OpenAI's encrypted content, summaries — as a `reasoning_details`
array beside the plaintext `reasoning`, and documents that the array must
go back verbatim and in order for a multi-turn tool round trip to keep the
model's continuity token; plaintext `reasoning_content` is only enough for
models that return raw reasoning strings. The adapter read the plaintext
and nothing else, so through OpenRouter a signed or encrypted block was
dropped on decode and the next request went out without it. The gateway
accepts that request and the model continues without its prior reasoning;
nothing errors, which is why it goes unnoticed.

Each entry is now carried as a `Custom` part, verbatim and tagged with the
model, the same way an unrecognised Anthropic block is; on the request
side, `Custom` parts whose `type` starts with `reasoning.` go out again as
`reasoning_details`, in order, beside the existing `reasoning_content`
echo (OpenRouter accepts both together — verified live against
`anthropic/claude-sonnet-4-6` and `moonshotai/kimi-k3`). Other `Custom`
parts stay ignored on this wire as before. Non-streaming path only; the
streaming `delta.reasoning_details` chunks are a follow-up.

Tests: the response path captures both entry types as `Custom` parts ahead
of the tool call, tagged with the model; the request path echoes the
entries verbatim beside `reasoning_content`; a non-reasoning `Custom` part
is still not echoed.

```

jeremychone:
@bricef Thanks, all good.

PR #302, opened 2026-09-07
fix(gemini) - keep a thoughtSignature next to the part it arrived on

### LINKED ISSUE #299: bug - Gemini adapter moves a `thoughtSignature` off the part it arrived on, and `ToolCall.thought_signatures` is written but never sent
**Narrow cargo test:** `cargo test --lib gemini::adapter_impl::tests::signatures_stay_next_to_the_part_they_arrived_on` (on the fork branch; fails on main)
**Adapter:** Gemini
**Model:** `gemini-3.5-flash`, `gemini-3-flash-preview` (any thinking model that signs function calls)

## What happens

Gemini attaches a `thoughtSignature` to one specific part of a response and wants it back on that same part. Observed live on 2026-09-07 (AI Studio, non-streaming, `includeThoughts: true`; the placement is the same without it):

| turn | parts returned |
|---|---|
| two parallel function calls | `thought`, `functionCall` **with** signature, `functionCall` without |
| visible text, then a function call | `thought`, `text` without signature, `functionCall` **with** signature |
| text only | `thought`, `text` **with** signature |

At `main` (`77894f7`) the non-streaming response path (`gemini/adapter_impl.rs`, `to_chat_response`) buckets the parsed content into signatures, texts, binaries and tool calls, then emits every `ThoughtSignature` first, then the merged text, then the tool calls, and mirrors all signatures onto the first tool call. The association between a signature and its part is lost at that point. The request path (`into_gemini_request_parts`, assistant branch) then attaches a pending signature to the next part it meets, and never reads `ToolCall.thought_signatures`. Consequences:

1. **Text followed by a signed call.** Decode yields `[signature, text, call]`. On the way back the signature is flushed as a bare `{"thoughtSignature": …}` part ahead of the text, and the call goes out with the `skip_thought_signature_validator` stand-in instead of the signature it was issued with. Both models above accepted that request — the stand-in exists to bypass the validator — so nothing errors: the turn continues with the signature somewhere it was not issued and the call carrying the bypass token rather than the model's state.
2. **A turn rebuilt from tool calls alone** (`ChatMessage::from(Vec<ToolCall>)`, the `c20-tooluse` example's pattern and what `ToolCall.thought_signatures` exists for) sends no signature at all: the stand-in on the first call for a `gemini-3*` model name, nothing for any other name.
3. **Two signed calls in one turn** would come back as `[sig-a, sig-b, call-a, call-b]` with both mirrored onto `call-a`, and be sent with `sig-a` bare, `sig-b` on `call-a` and nothing on `call-b`. The two models above sign only the first of a set of parallel calls, so this shape was not observed live; it is pinned with a hand-built body because it goes through the same code path.

Parallel calls with a single signature on the first call round-trip correctly today, by construction: the hoisted signature lands on the first call again.

## Proposed change

Keep wire order on decode, with each `ThoughtSignature` part immediately before the part it arrived on and each function call mirroring its own signature in `thought_signatures` (consecutive text parts still merge; thought summaries still go to `reasoning_content`). On encode, embed a signature in the text or `functionCall` part it precedes — the shape the API returned it in — and, when no signature part precedes a call, send the call's own `thought_signatures` before falling back to the stand-in. Non-streaming path; the streamer's end-of-stream assembly is a separate question.

A PR implementing this, with unit tests for decode order, per-call replay, text-part replay, the mirror, and a two-call round trip, is ready: `bricef/rust-genai` branch `fix/gemini-thought-signature-adjacency`. The existing recorded-cassette suites (`tests_yakbak_gemini`, `_gemini_ix`, `_anthropic`, `_openai`) pass unchanged.

## Verified live (2026-09-07, `gemini-3-flash-preview`, non-streaming)

With the branch, the parallel turn decodes as `[signature, call(get_weather, mirrored), call(get_local_time)]` and the text-then-call turn as `[text, signature, call(mirrored)]`. Each was replayed through `ChatMessage::assistant(response.content)` plus the tool responses, and both continuations were accepted, the model answering from the tool results. Raw replays of the API's own shapes were accepted too, as was the pre-change wire for the text-then-call turn — hence "moves", not "breaks".

bricef:
Fixes #299.

Gemini attaches a `thoughtSignature` to a specific part and wants it back on that same part: on the function call when there is one (only the first of a set of parallel calls carries one), on the last text part otherwise. The response path hoisted every signature to the front of the content and mirrored all of them onto the first tool call, so which part a signature belonged to was lost; the request path then attached a pending signature to the next part it met, so a text-plus-call turn sent its signature as a bare part ahead of the text and the call went out with the validator stand-in. `ToolCall.thought_signatures` was written on decode and never read on encode, so a turn rebuilt from tool calls alone (the `c20-tooluse` pattern) never sent a signature.

**Change.** The response keeps wire order: each signature part sits immediately before the part it arrived on, and each function call mirrors its own signature. Consecutive text parts still merge; thought summaries still go to `reasoning_content`. On the way back a signature is embedded in the text or `functionCall` part it precedes — the shape the API returned it in — and a call that carries a signature only in its mirror sends that rather than the Gemini 3 validator stand-in. A bare trailing signature is still sent as its own part.

**Behaviour note for reviewers.** Part order changes for callers who inspected `content.parts()` of a response with text before a signed call: the signature now sits between the text and the call rather than first. `ToolCall.thought_signatures` now holds one signature per call rather than all of them on the first.

**Verified live** on 2026-09-07 against `gemini-3-flash-preview`, non-streaming: a parallel two-call turn decodes as `[signature, call(mirrored), call]` and a text-then-call turn as `[text, signature, call(mirrored)]`; each replayed via `ChatMessage::assistant(response.content)` plus tool responses and continued normally. On `gemini-3.5-flash` and `gemini-3-flash-preview` only the first parallel call carries a signature, so the two-signed-calls test pins the code path rather than an observed shape. The pre-change wire for the text-then-call turn is also accepted by the API, since the stand-in bypasses its validator — this is a fidelity fix, not a fix for a rejection.

**Tests.** `signatures_stay_next_to_the_part_they_arrived_on`, `each_call_goes_back_with_its_own_signature`, `a_signature_on_a_text_part_rides_the_text_part_back`, `a_mirrored_signature_on_the_call_is_honoured`, `a_two_call_turn_round_trips_with_both_signatures`. Four fail against the pre-change code; the recorded-cassette suites for Gemini, Gemini Interactions, Anthropic and OpenAI pass unchanged.

**Scope.** Non-streaming path; the streamer's end-of-stream assembly is unchanged and worth a look separately. CHANGELOG entry under Gemini.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
   - The Anthropic `-zero` model suffix is canonical, while `-none` remains a backward-compatible alias. Both map to `Zero` and are stripped.
   - `^` Preserve signed thinking blocks and thought signatures across streaming and non-streaming tool turns. Rebuilds thinking blocks ahead of text and tool calls, adds `ChatResponse::into_assistant_message_for_tool_use`, and safely omits unpaired thinking blocks. (PR #275)
 - Gemini:
+  - `-` Keep a `thoughtSignature` next to the part it arrived on. The response keeps wire order instead of hoisting every signature to the front (consecutive text parts still merge), each function call mirrors its own signature in `thought_signatures` rather than the first call carrying all of them, and on the way back a signature is embedded in the text or `functionCall` part it precedes, with `ToolCall.thought_signatures` honoured when no signature part precedes the call. Fixes a two-call turn whose second call lost its signature and whose first call carried the wrong one, and a text-plus-call turn whose signature moved onto the text.
   - `^` Map `ReasoningEffort::Zero` to a budget of `0`, which might be rejected by the provider on some models.
 - OpenAI and Bedrock:
   - `^` Apply the `ReasoningEffort::Zero` rename to OpenAI and Bedrock adapter mappings while preserving provider-specific keyword mappings.
--- a/src/adapter/adapters/gemini/adapter_impl.rs
+++ b/src/adapter/adapters/gemini/adapter_impl.rs
@@ -189,55 +189,65 @@ impl Adapter for GeminiAdapter {
 		} = gemini_response;
 		let stop_reason = stop_reason.map(StopReason::from);
 
-		let mut thoughts: Vec<String> = Vec::new();
-		let mut reasonings: Vec<String> = Vec::new();
-		let mut texts: Vec<String> = Vec::new();
-		let mut tool_calls: Vec<ToolCall> = Vec::new();
-		let mut binary_parts: Vec<Binary> = Vec::new();
+		// Parts keep the order the API returned them in, and a thought signature
+		// stays immediately before the part it arrived on: Gemini attaches a
+		// signature to a specific function-call (or text) part and wants it back
+		// on that same part, so hoisting every signature to the front lost which
+		// part each belonged to, and a turn with two calls came back with both
+		// signatures on the first. Each call also mirrors its own signature in
+		// `thought_signatures`. Consecutive text parts still merge into one, as
+		// before; thought summaries go to `reasoning_content`.
+		let mut parts: Vec<ContentPart> = Vec::new();
+		let mut reasoning_text = String::new();
+		let mut text_buffer = String::new();
+		let mut pending_signature: Option<String> = None;
+
+		fn flush_text(text_buffer: &mut String, parts: &mut Vec<ContentPart>) {
+			if !text_buffer.is_empty() {
+				parts.push(ContentPart::Text(std::mem::take(text_buffer)));
+			}
+		}
 
 		for g_item in gemini_content {
 			match g_item {
-				GeminiChatContent::Text(text) => texts.push(text),
-				GeminiChatContent::Binary(binary) => binary_parts.push(binary),
-				GeminiChatContent::ToolCall(tool_call) => tool_calls.push(tool_call),
-				GeminiChatContent::ThoughtSignature(thought) => thoughts.push(thought),
-				GeminiChatContent::Reasoning(reasoning_text) => reasonings.push(reasoning_text),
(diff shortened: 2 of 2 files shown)


jeremychone:
Good fix/enhancement

- As you noted, we will have to do a follow-up on the streaming side. 
- Will also need `gemini_ix` to align with the new Google Interactions API with the same signature adjacency.

jeremychone:
PR Commit message

```
Gemini attaches a `thoughtSignature` to a specific part — a function
call, or a text part — and wants it back on that same part. The response
path hoisted every signature to the front of the content and mirrored all
of them onto the first tool call, so a turn with two signed calls lost
which signature belonged to which; the request path then attached a
pending signature to the next part it met, so the second call went out
without one (or with the first's), and a text-plus-call turn sent its
signature ahead of the text as a bare part. `ToolCall.thought_signatures`
was written on decode and never read on encode.

Now the response keeps wire order: each signature part sits immediately
before the part it arrived on, and each function call mirrors its own
signature in `thought_signatures`. Consecutive text parts still merge, and
thought summaries still go to `reasoning_content`. On the way back a
signature is embedded in the text or `functionCall` part it precedes — the
shape the API returned it in — and a call that carries a signature only in
its mirror sends that rather than Gemini 3's validator stand-in. A bare
trailing signature is still sent as its own part.

Tests: decode keeps order and mirrors per call; each call goes back with
its own signature; a signature on a text part rides the text part back;
a mirrored signature is honoured; a two-call turn round-trips whole.
Non-streaming path; the streamer's end-of-stream assembly is unchanged.

```

PR #303, opened 2026-09-07
fix(anthropic) - map output_tokens_details.thinking_tokens to completion_tokens_details.reasoning_tokens

### LINKED ISSUE #300: bug - anthropic: `usage.output_tokens_details.thinking_tokens` is not mapped to `completion_tokens_details.reasoning_tokens`
**What happens.** Anthropic's Messages API reports the thinking share of `output_tokens` as `usage.output_tokens_details.thinking_tokens` (adaptive thinking; seen on `claude-opus-5` with `anthropic-version: 2023-06-01`). The Anthropic adapter's `into_usage` sets `completion_tokens_details: None` (the comment says "for now, None for Anthropic"), so `ChatResponse.usage.completion_tokens_details` is always `None` for Anthropic — while the OpenAI adapter reads `completion_tokens_details.reasoning_tokens` and the Gemini adapter maps `thoughtsTokenCount` into the same field. A caller that splits thought from spoken tokens works for two adapters and silently gets nothing from the third.

**Observed** (live, 2026-09-06, non-streaming):

```json
"usage": {"input_tokens": 485, "output_tokens": 189, "output_tokens_details": {"thinking_tokens": 182}, "service_tier": "standard"}
```

→ `response.usage.completion_tokens_details == None`.

**Expected.** `completion_tokens_details.reasoning_tokens == Some(182)`, with `completion_tokens` still `Some(189)`: Anthropic's `output_tokens` already includes the thinking share, so unlike the OpenAI normalisation there is nothing to re-sum. When the field is absent the details should stay `None` rather than become a zero.

**Narrow cargo test:** `cargo test -p genai --lib anthropic::adapter_shared_tests::test_non_stream_usage_maps_thinking_tokens_to_reasoning_tokens` (on the linked PR; red without the change).

**Adapter:** Anthropic, non-streaming path. The streamer's `message_delta.usage` handling is a separate question — whether Anthropic sends `output_tokens_details` there is not verified here.

**Model:** `claude-opus-5`; any adaptive-thinking model.

**Version:** `0.7.0-beta.21` (crates.io) and `main` at `77894f7`.

bricef:
Fixes #300.

Anthropic reports the thinking share of `output_tokens` as `usage.output_tokens_details.thinking_tokens`, and `into_usage` dropped it (`completion_tokens_details: None`, "for now"). It now lands in `completion_tokens_details.reasoning_tokens`, the field the OpenAI and Gemini adapters already fill, so the thought-versus-spoken split reads the same way across the three adapters.

**Change.** `into_usage` reads `output_tokens_details.thinking_tokens` when present and builds a `CompletionTokensDetails` with only `reasoning_tokens` set. `completion_tokens` is untouched: Anthropic's `output_tokens` already includes the thinking share, unlike OpenAI's `completion_tokens`, so there is no re-summing. When the field is absent the details stay `None` — nothing is invented.

**Tests.** `test_non_stream_usage_maps_thinking_tokens_to_reasoning_tokens` (fails without the change) and `test_non_stream_usage_without_thinking_share_has_no_completion_details`. The existing Anthropic suites pass unchanged.

**Scope.** Non-streaming path. The streamer's `message_delta.usage` handling is unchanged; whether Anthropic sends `output_tokens_details` on that path is not verified here. CHANGELOG entry under Anthropic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
   - Fable and Mythos omit `thinking`, because it is always on and cannot be explicitly disabled.
   - The Anthropic `-zero` model suffix is canonical, while `-none` remains a backward-compatible alias. Both map to `Zero` and are stripped.
   - `^` Preserve signed thinking blocks and thought signatures across streaming and non-streaming tool turns. Rebuilds thinking blocks ahead of text and tool calls, adds `ChatResponse::into_assistant_message_for_tool_use`, and safely omits unpaired thinking blocks. (PR #275)
+  - `-` Map `usage.output_tokens_details.thinking_tokens` to `completion_tokens_details.reasoning_tokens` on the non-streaming path, so the thinking share of `output_tokens` reads the same way as for the OpenAI and Gemini adapters; `completion_tokens` is unchanged, since Anthropic's `output_tokens` already includes it.
 - Gemini:
   - `^` Map `ReasoningEffort::Zero` to a budget of `0`, which might be rejected by the provider on some models.
 - OpenAI and Bedrock:
--- a/src/adapter/adapters/anthropic/adapter_shared.rs
+++ b/src/adapter/adapters/anthropic/adapter_shared.rs
@@ -6,8 +6,9 @@ use crate::adapter::adapters::support::get_api_key;
 use crate::adapter::{Adapter, AdapterKind, ServiceType, WebRequestData};
 use crate::chat::{
 	Binary, BinarySource, CacheControl, CacheCreationDetails, ChatOptionsSet, ChatRequest, ChatResponse,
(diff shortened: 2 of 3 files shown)


jeremychone:
Thanks, looks good!

A minor followup PR will be made to treat zero as `None` for `thinking_tokens`, since this is the way it is done for all other usage counters across the codebase.

jeremychone:
PR commit message

```
Anthropic reports the thinking share of `output_tokens` as
`usage.output_tokens_details.thinking_tokens`, and `into_usage` dropped it
(`completion_tokens_details: None`, "for now"). It now lands in
`completion_tokens_details.reasoning_tokens`, the field the OpenAI and
Gemini adapters already fill, so the thought-versus-spoken split reads
the same way across the three adapters. `completion_tokens` is untouched:
Anthropic's `output_tokens` already includes the thinking share, so
there is no re-summing. Absent field, absent details.

```

PR #304, opened 2026-09-08
pr-followup for #296 and #297 (remove panicables)

jeremychone:
## Summary

This follow-up addresses maintainer feedback for PRs #296 and #297.

## Changes

- Removed panic-prone `unwrap()` and `expect()` calls from the Bedrock streaming parser and related tests.
- Replaced unchecked Bedrock frame and header slicing with explicit bounds validation and fallible access.
- Preserved all events decoded from a single Bedrock frame by queueing pending events.
- Replaced the OpenAI Responses replay test panic with fallible error propagation.
- Replaced the remaining test-only `unwrap()` in the OpenAI Responses streamer.
- Added `// -- ...` section comments to delimit logic blocks in the changed functions and tests.

## Validation

The affected streaming paths now use fallible handling for malformed input and unexpected test states instead of panicking.


jeremychone:
@emecii So, I made this change, and it should not affect the logic, but feel free to submit another PR if something is wrong.

PR #305, opened 2026-09-08
. pr-followup #303 - usage details zero-as-none

jeremychone:
# PR - . pr-followup #303 - usage details zero-as-none

Follow-up to PR #303 (issue #300).

## Summary

In PR #303, `usage.output_tokens_details.thinking_tokens` was mapped to `completion_tokens_details.reasoning_tokens` for the Anthropic non-streaming adapter path.

This follow-up ensures that if Anthropic returns `output_tokens_details.thinking_tokens` with a count of `0`, it is treated as `None` rather than `Some(0)`. This aligns with genai's zero-as-none normalization convention across usage details (such as `PromptTokensDetails` and `zero_as_none` deserializers), ensuring an empty `CompletionTokensDetails` struct is not instantiated when no thinking tokens were used.

## Changes

- In `src/adapter/adapters/anthropic/adapter_shared.rs`, filtered out non-positive values via `.filter(|&tokens| tokens > 0)` when extracting `thinking_tokens`.
- In `src/adapter/adapters/anthropic/adapter_shared_tests.rs`, added `test_non_stream_usage_zero_thinking_tokens_is_none`.

### DIFF
--- a/src/adapter/adapters/anthropic/adapter_shared.rs
+++ b/src/adapter/adapters/anthropic/adapter_shared.rs
@@ -79,7 +79,8 @@ impl AnthropicAdapter {
 			.get("output_tokens_details")
 			.and_then(|details| details.get("thinking_tokens"))
 			.and_then(Value::as_i64)
-			.map(|tokens| tokens as i32);
+			.map(|tokens| tokens as i32)
+			.filter(|&tokens| tokens > 0);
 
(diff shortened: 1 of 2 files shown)


jeremychone:
@bricef Here is the quick follow-up PR I did, to keep the `none as zero` as others. 

Feel free to chime in if you think it is the wrong behavior.

bricef:
Makes sense to keep this consistent @jeremychone 👍

PR #308, opened 2026-09-11
Pr/bedrock aws fixes

wdwind:
Three independent fixes for the AWS Bedrock adapters. Each has a regression test that fails without it.

**1. `bedrock_sigv4` — refresh credentials before they expire.**
`provide_credentials()` returns an immutable snapshot, but it was cached in a `OnceCell` for the life of the process. Temporary credentials (STS, IMDS, SSO) expired roughly an hour in, after which every request failed with a signature error until restart — unrecoverable, because the cache was a private `static` and this adapter discards `AuthData`. The cache now mirrors the AWS SDK's identity cache: treat credentials as expired 10 seconds early, fall back to a 15-minute TTL when the provider reports no expiry, deduplicate concurrent refreshes, and jitter them.

**2. `bedrock_api` — accept `AWS_BEARER_TOKEN_BEDROCK`.**
Only `BEDROCK_API_KEY` was read, while AWS documents `AWS_BEARER_TOKEN_BEDROCK` — the name the AWS SDKs use — so following the AWS docs silently found no key. Both are accepted now, `BEDROCK_API_KEY` first, and an empty value falls through instead of shadowing the other. An explicit `AuthData` (key, custom env var, or resolver result) is still used as-is.

**3. `bedrock_sigv4` — build the request URL for the region we sign for.**
Two independent lookups fed one signed request: the endpoint came from `resolve_region()` (env only, defaulting to `us-east-1`), while the SigV4 credential scope came from `aws-config`, which also reads the profile region in `~/.aws/config` and IMDS. With a profile-only region, the request went to the `us-east-1` host signed for `eu-west-1`, and AWS answered `SignatureDoesNotMatch`. The reconciling `override_endpoint_region` matched a literal `bedrock-runtime..amazonaws.com` placeholder that no code path produces, so it never ran; it now rebuilds the endpoint when it recognises its own env-derived default, and leaves user-supplied endpoints (VPC endpoint, proxy, gateway) alone.

**Tests.** New unit tests cover the refresh deadline/buffer/jitter, the env-var precedence, and the URL/signing-region alignment — the last one fails against the previous implementation.

**Scope.** 6 files, Bedrock adapters only. The `bedrock_sigv4` `block_in_place` runtime requirement is unchanged.

Validation:

- `cargo test --features bedrock-sigv4 --lib` — 335 passed
- `cargo check --features bedrock-sigv4 --all-targets`
- `rustfmt --check` on the changed files
- `git diff --check`

### DIFF
--- a/examples/c99-bedrock-api.rs
+++ b/examples/c99-bedrock-api.rs
@@ -4,10 +4,13 @@
 //! and no extra dependencies pulled in. It authenticates with a Bedrock API key passed in the
 //! `Authorization: Bearer` header.
 //!
-//! Required env vars:
-//!   - BEDROCK_API_KEY: a Bedrock API key (see
+//! Required env var — either of these names works:
+//!   - BEDROCK_API_KEY, OR
+//!   - AWS_BEARER_TOKEN_BEDROCK (the name AWS documents, see
 //!     https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html)
-//!   - AWS_REGION (optional): defaults to `us-east-1`. Model availability varies by region.
+//!
+//! Optional:
+//!   - AWS_REGION: defaults to `us-east-1`. Model availability varies by region.
 //!
 //! Run with: `cargo run --example c99-bedrock-api`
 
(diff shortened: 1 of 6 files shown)


jeremychone:
@wdwind Thanks for this good PR.

PR #310, opened 2026-09-13
+ bedrock_sigv4 - select the AWS profile per client via AuthData

wdwind:
## What

`bedrock_sigv4` now reads `AuthData` as an AWS profile name, so a caller can choose the account per client:

    let dev = Client::builder()
        .append_provider_config(AdapterKind::BedrockSigv4, AuthData::Key("dev".to_string()))
        .build()?;

`Key("<profile>")`, `FromEnv("<VAR>")`, and `MultiKeys({ "profile": "<name>" })` all work. `None` or a blank value keeps today's behavior (`AWS_PROFILE`, else `default`); `RequestOverride` is untouched.

Previously the profile came from `AWS_PROFILE`, resolved once per process, so two clients in one process could not use different profiles.

The process-wide credential cache became a per-profile cache: each profile keeps its own provider, region, and credentials, and still refreshes them before expiry. 

Tests cover profile extraction and the cache (fresh reuse, refresh in place, one profile's fetch not blocking another), plus a live smoke test in `tests/tests_p_bedrock_sigv4.rs`. `cargo test --features bedrock-sigv4 --lib`, clippy under the same features, and `cargo test --features "bedrock-sigv4,otel" --no-run` are green.

## Notes

- Includes one incidental fix: `Error::ClientBuildFail` was missing from the OTel `error.type` mapping,
  so `otel` + `bedrock-sigv4` did not compile.

### DIFF
--- a/docs/for-llm/api-reference-for-llm.md
+++ b/docs/for-llm/api-reference-for-llm.md
@@ -38,7 +38,7 @@ genai (crate root / lib.rs)
   - `vertex` (since v0.6.0) is Google Vertex AI Model Garden routing for Gemini and Anthropic models via the `vertex::` namespace.
   - `opencode_go` (since v0.6.0) is a curated open coding model proxy with dual-protocol routing via the `opencode_go::` namespace. Uses `OPENCODE_GO_API_KEY`.
   - `bedrock_api` (since v0.6.0) is AWS Bedrock Converse API, authenticated with a simple Bearer token from `BEDROCK_API_KEY`. Use via the `bedrock_api::` namespace.
-  - `bedrock_sigv4` (since v0.6.0) is AWS Bedrock Converse API, authenticated via SigV4 + standard AWS credential chain. Requires the `bedrock-sigv4` feature. Use via the `bedrock_sigv4::` namespace.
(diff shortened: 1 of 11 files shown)


jeremychone:
PR Commit message:

```
- `AuthData::Key("<profile>")`, `FromEnv("<VAR>")`, and `MultiKeys({ "profile": .. })` now select the AWS profile; `None`/blank keeps the ambient chain (`AWS_PROFILE`, else `default`). `RequestOverride` behavior is unchanged.
- AWS config, region, and credentials are resolved and cached per profile instead of process-wide, so clients using different AWS accounts can coexist in one process.
- Share the region fallback across both Bedrock adapters, and document the profile-name semantics of `AuthData` on the enum itself.
- Add cache-behavior tests (fresh reuse, refresh in place, per-profile isolation, and one profile loading without blocking another) plus a live smoke test (tests_p_bedrock_sigv4).
- Map `Error::ClientBuildFail` in the OTel `error.type` derivation, which the `otel` + `bedrock-sigv4` combination needed to compile.
```

jeremychone:
@wdwind Thank you for this PR. During my review with AI, it noticed a comment that was out of sync and recommended adding the following to sigv4.rs, which I did in a subsequent commit.

`sigv4.rs` added comment: 
```
//! Unlike the SDK, concurrent refreshes for the same profile are not deduplicated;
//! the lock is never held across a fetch, so one profile's slow fetch does not block another.
//! Without the cache, a long-lived process would keep signing with credentials that expired an hour after start-up
//! (`provide_credentials()` returns a frozen snapshot).
```

If this is not correct, feel free to open a small PR to address it.

PR #311, opened 2026-09-14
fix(ollama): emit every event of an ndjson line instead of dropping the extras

dmfs:
- poll_next used to return on the first event-bearing field of a line, so fields sharing a line (e.g., `thinking` + `content`), all but the first tool call of a line, and all lines after the first event-bearing line of a web chunk were silently dropped

- now each line is parsed into a buffered event queue (reasoning, then content, then one ToolCallChunk per tool call) that poll_next drains in stream order before polling the inner stream again

- the End event is queued as well, so events of the final `done` line are emitted first and the captures of that line (usage, done_reason) are no longer lost; lines after the `done` line are ignored

- OllamaStreamer becomes generic over the inner stream (Stream<Item = Result<String, BoxError>>) so unit tests can feed raw ndjson lines; new tests cover combined-field lines, multi-line chunks, event ordering, End captures, and StreamParse errors


jeremychone:
pr commit message:
```
- poll_next used to return on the first event-bearing field of a line, so fields sharing a line (e.g., `thinking` + `content`), all but the first tool call of a line, and all lines after the first event-bearing line of a web chunk were silently dropped

- now each line is parsed into a buffered event queue (reasoning, then content, then one ToolCallChunk per tool call) that poll_next drains in stream order before polling the inner stream again

- the End event is queued as well, so events of the final `done` line are emitted first and the captures of that line (usage, done_reason) are no longer lost; lines after the `done` line are ignored

- OllamaStreamer becomes generic over the inner stream (Stream<Item = Result<String, BoxError>>) so unit tests can feed raw ndjson lines; new tests cover combined-field lines, multi-line chunks, event ordering, End captures, and StreamParse errors
```

PR #312, opened 2026-09-14
feat(ollama): map ReasoningEffort to the think body param

dmfs:
- `think` is a top-level Ollama body param (like `format`, not inside `options`) and accepts a boolean or a level string ("low" | "medium" | "high" | "max")

- mapping: Zero -> false, Minimal/Low -> "low", Medium -> "medium", High -> "high", XHigh/Max -> "max", Budget(_) -> true

- Minimal maps to "low" and XHigh/Max to "max" since Ollama has no equivalent levels; Budget enables thinking at the model default level because Ollama has no token-budget knob

- `think` is omitted when no reasoning_effort is set

- add payload unit tests for the mapping and a live qwen3 test asserting that ReasoningEffort::Zero disables thinking (no reasoning_content)

Ref: https://docs.ollama.com/api/chat#body-think-one-of-0


ropoctl:
this fixes an adapter, but it seems like there should be a yakbak cassette recording + test using it in here if this changes behavior of an adapter

jeremychone:
@dmfs Thank you for this PR. Can you let me know if a Yakbak/Cassette recording would be needed. 

@ropoctl added this Yakbak/Cassette testing scheme, and it helps reliability a lot.

dmfs:
Please correct me if I'm wrong. As far as I can see the yakbak implementation ignores the request and only replays the response. Since the change is solely in the request, another yakbak test for Ollama wouldn't really add any benefit. Am I missing something?

ropoctl:
My apologies, i must have intended this comment to go with your other pr, that was merged today, would you be willing to follow up #311

dmfs:
okay, I'll take care of that. Shouldn't be a blocker for this PR I suppose.

dmfs:
see #313

jeremychone:
PR commit message

```
- `think` is a top-level Ollama body param (like `format`, not inside `options`) and accepts a boolean or a level string ("low" | "medium" | "high" | "max")

- mapping: Zero -> false, Minimal/Low -> "low", Medium -> "medium", High -> "high", XHigh/Max -> "max", Budget(_) -> true

- Minimal maps to "low" and XHigh/Max to "max" since Ollama has no equivalent levels; Budget enables thinking at the model default level because Ollama has no token-budget knob

- `think` is omitted when no reasoning_effort is set

- add payload unit tests for the mapping and a live qwen3 test asserting that ReasoningEffort::Zero disables thinking (no reasoning_content)

Ref: https://docs.ollama.com/api/chat#body-think-one-of-0
```

jeremychone:
@dmfs Thanks

PR #313, opened 2026-09-16
. tests - add yakbak ollama ndjson multi-event replay tests

dmfs:


jeremychone:
@dmfs My AI reviewing step gave me that (deepseek-flash-v4.1)

```
Minor (non-blocking) notes:

- The event-order test uses `while let Some(Ok(event)) = stream.next().await`, which silently swallows a stream error and stops the loop. The final `assert_eq!` would still fail, but the actual error detail is lost. Using the `extract_stream_end` helper (or surfacing the error) would keep the failure informative.
- Inconsistent import style between the two tests: `serde_json::json` and the support imports are at the top, while `tokio_stream::StreamExt` is imported inside the second test body. Not a defect, just a small consistency polish.
- The cassette file has no trailing newline ("No newline at end of file"). Harmless for the replay reader, since each line is parsed independently.
```

Not sure if this an issue or not. Might be false positive, I let you decide if any action should be taken, but this PR was merged.

dmfs:
thx, I'll feed it to my agent and ask it validate and fix these :-)

PR #315, opened 2026-09-16
Bedrock OpenAI reasoning effort

kaeluka:
Hi - reasoning levels for openai models on bedrock can't be controlled. This PR works for me :-)

I'd appreciate a review.

And thanks for the great library!

---

## What changed

Bedrock's Converse adapter now forwards an explicitly-set `ReasoningEffort` for OpenAI models instead of ignoring it.

- `BedrockPublisher` gains `OpenAI` and `OpenAIGptOss` variants. `from_model_id` recognizes `openai.<model>` IDs, including the cross-region prefixes `us.`/`eu.`/`apac.`/`global.`/`in.`, and classifies `gpt-oss-*` separately from other OpenAI models. Matching is on the publisher segment only, so custom IDs that merely contain `openai` are not reclassified.
- `publisher_additional_fields` maps `ReasoningEffort` to Bedrock's OpenAI keywords:
  - `gpt-oss-*` -> flat `additionalModelRequestFields.reasoning_effort` (documented by AWS for GPT-OSS; accepted live on Bedrock via OpenRouter with reasoning output)
  - other OpenAI models -> nested `additionalModelRequestFields.reasoning.effort` (observed live for GPT-5.6 luna/terra/sol; **inferred for gpt-5.4/5.5 and gpt-6-astra, not yet observed**)
  - `Budget(_)` and unset effort are omitted; `Zero` maps to `"none"`.
- `resolve_max_tokens` gives OpenAI models a 4096 default, matching `Other` (no behavior change vs. before).

## Why

There was no way to control reasoning effort for OpenAI models on Bedrock: the Converse adapter only populated `additionalModelRequestFields` for Anthropic/Nova. OpenAI's Bedrock models expect two different field shapes depending on family (flat `reasoning_effort` for GPT-OSS vs. nested `reasoning.effort` for newer models), which is why the publisher enum distinguishes them.

## Verification

The two shapes come from live Converse/ConverseStream evidence: flat `reasoning_effort` for GPT-OSS, nested `reasoning.effort` for GPT-5.6 (vercel/ai#19403, vercel/ai#19420). gpt-5.4/5.5 and gpt-6-astra use the same family heuristic but haven't been exercised against a live endpoint; if Bedrock rejects the shape for one of them it fails loudly with a provider 400, never silently. Serialization is covered by signed-request tests on both shapes; direct AWS verification remains blocked by an account-level Error 002.

## Testing

Unit tests cover both Bedrock adapters, all profile prefixes, the two field shapes, omission for unset/`Budget`, unchanged Anthropic/Nova/other mappings, and non-reclassification of custom IDs. `cargo test --lib` (335 passed) and `cargo test --lib --features bedrock-sigv4` (354 passed).


jeremychone:
@kaeluka In the description of this PR, can you describe concisely what has changed and why, so we capture the intent?

kaeluka:
@jeremychone done! 🙂

jeremychone:
Thanks @kaeluka, nice work, and thanks for the clear description.

I did a first crontrolled AI review of this PR, and here is what it gave me. consensus of Flash, Opus and Sol. 

Review it, make any changes you think are needed to address those issues, and let me know if you disagree with anything.

You can definitely use AI to review and write the code, as long as you carefully review the code and leave a comment before pushing or updating.

Again, big thanks. This seems to be a good emprovements. I'm being a little overcautious here because I don't run those models/providers on my side, so I want to make sure we don't break anything for anybody else.

AI review below:
----

Before merging, a few changes are needed. `from_model_id` is shared by all Bedrock models, so this PR also changes behavior for Anthropic and Nova, not just OpenAI.



### 1. Profile prefix regression (blocker)

The old logic always used the second segment, so any inference-profile prefix worked. The new code only strips `us`, `eu`, `apac`, `global`, `in`.

IDs like `jp.anthropic...`, `au.anthropic...`, `us-gov.anthropic...`, and system inference-profile ARNs now resolve to `Other`, which:

- drops the default Claude `maxTokens` from 64000/32000 to 4096 (silent truncation)
- drops Anthropic `thinking` and Nova reasoning fields

Suggested fix: use the first segment when it is a known publisher, otherwise fall back to the second segment. This keeps your direct-ID fix.

```rust
pub(super) fn from_model_id(model_id: &str) -> Self {
	// Profile ARNs end with the profile ID after the last '/'.
	let id = model_id.rsplit_once('/').map(|(_, id)| id).unwrap_or(model_id);
	let mut segments = id.split('.');
	let first = segments.next().unwrap_or_default();
	// Direct IDs start with the publisher; any other leading segment is an inference-profile prefix.
	let publisher = if matches!(first, "anthropic" | "amazon" | "openai") {
		first
	} else {
		segments.next().unwrap_or_default()
	};

	match publisher {
		"anthropic" => Self::Anthropic,
		"amazon" => Self::AmazonNova, // Nova models; Titan would also hit this
		"openai" if segments.next().unwrap_or_default().starts_with("gpt-oss-") => Self::OpenAIGptOss,
		"openai" => Self::OpenAI,
		_ => Self::Other,
	}
}
```

Note: with this, `custom.openai.gpt-5.6-luna` resolves to `OpenAI` (same as the old logic). It is not a real Bedrock ID, so please drop that assertion. `us.custom.openai...` and `custom-openai.gpt-oss...` still resolve to `Other`.

### 2. Clamp GPT-OSS keywords

GPT-OSS only accepts `low`, `medium`, `high`. A user with a client-wide default of `Zero`, `XHigh`, or `Max` had a working request before (effort ignored) and would now get a 400.

```rust
BedrockPublisher::OpenAIGptOss => {
	let keyword = match effort {
		ReasoningEffort::Minimal | ReasoningEffort::Low => "low",
		ReasoningEffort::Medium => "medium",
		ReasoningEffort::High | ReasoningEffort::XHigh | ReasoningEffort::Max => "high",
		// GPT-OSS cannot disable reasoning and has no token budget.
		ReasoningEffort::Zero | ReasoningEffort::Budget(_) => return None,
	};
	Some(json!({ "reasoning_effort": keyword }))
}
```

Keep your full mapping for the nested `reasoning.effort` shape. It matches the native OpenAI adapter (including `Max` -> `"max"`).

### 3. Tests

Please add regression tests asserting both `maxTokens` and `additionalModelRequestFields` for:

- direct and profile `anthropic.*` and `amazon.*` IDs
- `jp.`, `au.`, and `us-gov.` prefixes
- a system inference-profile ARN

Optional, to match project conventions: `test_` prefix on test names, and `// -- Setup & Fixtures`, `// -- Exec`, `// -- Check` sections.

kaeluka:
> I'm being a little overcautious here because I don't run those models/providers on my side, so I want to make sure we don't break anything for anybody else.

I'm glad you are! I'm going to take a look ASAP.

